PR #10, opened 2020-08-31
Add 404 and 500 error pages

mattfredericksen:
Both pages use a slightly modified version of [this template](https://colorlib.com/etc/404/colorlib-error-404-4/). Credit is given in a comment.

Potential issues:
- Currently no context passing occurs. How much detail should users know about why an error occurred? (Exception messages are shown in debug mode.)
- `error.html`
  - Currently contains a bunch of css. Not sure if that should be moved elsewhere.
  - Located under `templates/`. Not sure if that's where it should stay. `404.html` and `500.html` must remain there.


templates/error.html
```diff
@@ -0,0 +1,139 @@
+{% extends 'base.htm' %}
+
+{% block content %}
```
iacobfred, line 3: Indentation is inconsistent; can you format the file?  PyCharm might be able to autoindent it for you.  Our standard is 4 spaces per indentation.

templates/error.html
```diff
@@ -0,0 +1,139 @@
...
+    <!-- This template was made by Colorlib (https://colorlib.com)
```
iacobfred, line 4: Can you remove the Colorlib comments?

templates/error.html
```diff
@@ -0,0 +1,139 @@
...
+         Please visit our website for more awesome templates, themes and tools. -->
+	<div id="notfound">
```
iacobfred, line 6: Since this is the base error template, let's choose more generic `id`s (to describe error elements rather than specifically 404 elements).

templates/error.html
```diff
@@ -0,0 +1,139 @@
...
+		<div class="notfound">
+			<div class="notfound-404">
+				<h1>Oops!</h1>
+				<h2>
+                    {% block message %}{% endblock %}
+                </h2>
+			</div>
+			<a href="/">Go TO Homepage</a>
```
iacobfred, line 14: Let's remove this link.  The logo in the navbar already functions as a link to the homepage, and if the user wants to go back to the previous page (before the error happened), they can just use the browser's back button.

templates/error.html
```diff
@@ -0,0 +1,139 @@
...
+		</div>
+	</div>
+{% endblock content %}
+
+{% block styles %}
+    <meta charset="utf-8">
```
iacobfred, line 20: Let's remove these meta tags.  

The `styles` block is just for CSS (although we could add another block in base.htm for additional meta tags, if needed). The `base.htm` template already has the `charset` and `viewport` meta tags:
```
<meta charset="UTF-8" />
<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
```
As for the http-equiv meta tag, it's probably not necessary, since we probably don't need to worry about supporting IE 8/9 (see https://stackoverflow.com/questions/6771258/what-does-meta-http-equiv-x-ua-compatible-content-ie-edge-do).

templates/error.html
iacobfred: Note: The CSS rules will need to be updated when the ids and classes are changed.

templates/error.html
mattfredericksen: The template is licensed under [CC BY 3.0](https://creativecommons.org/licenses/by/3.0/). I think we have to give credit somewhere, but it doesn't necessarily have to be in a comment. Or we could [buy the template](https://colorlib.com/wp/licence/) for $29. Or we could get rid of this template and make our own. 

Thoughts?

templates/error.html
mattfredericksen: I can update that. That was just what came with the Colorlib template.

templates/error.html
mattfredericksen: Ah, there were just some tabs thrown in there.

templates/error.html
iacobfred: Ah, I see.  Let's keep the attribution for now, then, and when we modify the styles and content sufficiently (to make the error page more cohesive with the rest of the site), we can get rid of it.

templates/error.html
mattfredericksen: Are `HTTP-error` and `HTTP-error-content` acceptable?

iacobfred (approved):

PR #222, opened 2020-12-15
implement ASGI

iacobfred:


mattfredericksen (approved):
Worker/thread count should be researched further.

PR #235, opened 2020-12-31
refactoring

iacobfred:


Dockerfile.frontend
```diff
@@ -19,8 +19,9 @@ RUN chown -R www-data:www-data /modularhistory && \
   chmod g+w -R /modularhistory/static && \
   chmod +x /modularhistory/config/wait-for-it.sh
 
-# Expose port 3000
+# Expose ports
 EXPOSE 3000
+EXPOSE 49153
```
mattfredericksen, line 24: We no longer need this port exposed here or mapped in [docker-compose.yml](https://github.com/ModularHistory/modularhistory/pull/235/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R15).

mattfredericksen (approved):

Dockerfile.frontend
iacobfred: I'll remove it in my next changeset.

PR #236, opened 2021-01-02
fix media URL references; remove unnecessary port exposure

iacobfred:


iacobfred:
@mattfredericksen, This should fix the image source URLs.

mattfredericksen (approved):

PR #246, opened 2021-01-08
add api views

iacobfred:
@mattfredericksen

### DIFF
--- a/apps/entities/api/views.py
+++ b/apps/entities/api/views.py
@@ -1,7 +1,6 @@
 from rest_framework import permissions
 from rest_framework.generics import ListAPIView
 from rest_framework.viewsets import ModelViewSet
-from modularhistory.views import AsyncAPIViewMixin
 from apps.entities.models import Entity
 from apps.entities.serializers import EntitySerializer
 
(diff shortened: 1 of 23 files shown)


apps/entities/api/views.py
```diff
@@ -0,0 +1,21 @@
...
+
+class EntityViewSet(ModelViewSet):
+    """API endpoint for viewing and editing entities."""
+
+    queryset = Entity.objects.exclude(type='entities.deity').order_by('birth_date')  # type: ignore
+    serializer_class = EntitySerializer
+    permission_classes = [permissions.IsAuthenticated]
+
+
+class EntityListAPIView(AsyncAPIViewMixin, ListAPIView):
```
mattfredericksen, line 17: Do we want to add the overhead of [AsyncAPIViewMixin](https://github.com/ModularHistory/modularhistory/blob/245/modularhistory/views.py#L7) to each of these without receiving any current benefit?

mattfredericksen (changes requested):

mattfredericksen (approved):

PR #248, opened 2021-01-11
enable Next.JS; refactoring

iacobfred:


docker-compose.yml
```diff
@@ -99,9 +75,9 @@ services:
       dockerfile: Dockerfile.react
     command: npm run dev
     depends_on:
-      - dev_backend
+      - django
     links:
-      - dev_backend:drf
+      - django:drf
```
mattfredericksen, line 80: @iacobfred we can remove this link now that the underscore is gone. The api link in [`frontend/pages/entities/index.js`](https://github.com/ModularHistory/modularhistory/blob/b4b65ab55d4a7557165867bae3c8027693bfc6c3/frontend/pages/entities/index.js#L104) will need to be updated. We should create a new environment variable for the api root (`http://django:8001/api/`).

mattfredericksen (commented):
My understanding is that mapping ports in [`docker-compose.yml`](https://github.com/ModularHistory/modularhistory/pull/248/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3) is only useful in development, where you might want to access `localhost:<port>`. In production, every container has access to the others through service names, right? Should the port mappings be moved to [`docker-compose.override.yml`](https://github.com/ModularHistory/modularhistory/pull/248/files#diff-2286cdeca85adb587a9640aa34352ba7c2fb6e49c0129dab05892ff67d0d4c58)?

iacobfred:
> My understanding is that mapping ports in [`docker-compose.yml`](https://github.com/ModularHistory/modularhistory/pull/248/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3) is only useful in development, where you might want to access `localhost:<port>`. In production, every container has access to the others through service names, right? Should the port mappings be moved to [`docker-compose.override.yml`](https://github.com/ModularHistory/modularhistory/pull/248/files#diff-2286cdeca85adb587a9640aa34352ba7c2fb6e49c0129dab05892ff67d0d4c58)?

@mattfredericksen, you're right that every container has access to the others through service names.  However, considering that we're not currently in a Docker-only environment in production, I think it's safest to keep the ports mapped.  As long as the host doesn't require those ports for other services, there's no harm in keeping them mapped.

mattfredericksen:
> (And as long as the host doesn't require those ports for other services, there's no harm in keeping them mapped.) 

@iacobfred my only real concern would be if mapping the ports made them visible to the internet. Even if requests sent to them are rejected, open ports are a security risk. Is the nginx process the only one which is actually exposed to the internet?

iacobfred:
> > (And as long as the host doesn't require those ports for other services, there's no harm in keeping them mapped.)
> 
> @iacobfred my only real concern would be if mapping the ports made them visible to the internet. Even if requests sent to them are rejected, open ports are a security risk. Is the nginx process the only one which is actually exposed to the internet?

Yep, 80 and 443 are the only ports exposed to the Internet.

iacobfred:
@mattfredericksen, ready for review!

mattfredericksen (approved):

PR #289, opened 2021-03-01
MH-11: Reactify auth

iacobfred:


.vscode/launch.json
```diff
@@ -14,6 +14,17 @@
             ],
             "envFile": "${workspaceFolder}/.env",
             "django": true
-        }
+        },
+        // https://marketplace.visualstudio.com/items?itemName=msjsdiag.debugger-for-chrome
```
iacobfred, line 18: Ah, need to remove this.

.config/nginx/dev/nginx.conf
```diff
@@ -67,17 +86,6 @@ server {
         auth_basic off;
     }
 
-    location ~ ^/(_?_next.*|entities/?$) {
```
iacobfred, line 70: Moved.

apps/users/views/auth.py
```diff
@@ -16,6 +27,10 @@ class LoginView(auth_views.LoginView):
     form_class = LoginForm
     template_name = 'users/login.html'
 
+    def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
+        """Respond to GET requests for the login page."""
+        return HttpResponsePermanentRedirect('/api/auth/signin')
```
mattfredericksen, line 32: I don't see any urls using this view in [`/modularhistory/urls.py`](https://github.com/ModularHistory/modularhistory/blob/MH-11/modularhistory/urls.py), [`/apps/users/urls.py`](https://github.com/ModularHistory/modularhistory/blob/MH-11/apps/users/urls.py), or [`/apps/users/api/urls.py`](https://github.com/ModularHistory/modularhistory/blob/MH-11/apps/users/api/urls.py). Do we foresee this view being used anywhere?

iacobfred: Oh, we can probably get rid of this. I don't foresee it being used, and it would be easy to recreate if necessary.

frontend/axios.ts
```diff
@@ -0,0 +1,10 @@
+import axios from 'axios';
+
+const djangoCsrfCookieName = 'csrftoken';
+
+// Axios config for DRF requests
+axios.defaults.xsrfHeaderName = 'X-CSRFToken';
+axios.defaults.xsrfCookieName = djangoCsrfCookieName;
+axios.defaults.withCredentials = true;
+
+export default axios;
```
mattfredericksen, line 10: Should we now import `axios` from this location every time we need it, or only when the `CSRFtoken` is needed?

iacobfred: Only when the CSRF token is needed (which means anytime that we are making a POST request to our Django API, I think).

frontend/pages/auth/signout.jsx
```diff
@@ -0,0 +1,65 @@
...
+      document.cookie = `next-auth.session-token=; path=/; expires=Thu, 01 Jan 1970 00:00:01 GMT`;
+    })
+    .catch(function (error) {
+      console.error(`Failed to sign out due to error:\n${error}`);
+      throw new Error(`${error}`);
+    });
+    // Sign out of the front end.
+    signOut();
+    // To trigger the event listener we save some random data into the `logout` key
+    window.localStorage.setItem("logout", Date.now()); // new
```
mattfredericksen, line 58: Could you clarify the purpose of this? What event listener needs to be triggered?

iacobfred: Ah, I forgot to finish implementing this. For now, I'll remove it.

It was intended to let other windows know to (immediately) log out, if you have log out in one window when you have multiple windows open.

frontend/next.config.js
```diff
@@ -1,32 +1,15 @@
...
-} else {
-  throw new Error(".env var 'HOSTNAME' is already defined");
-}
+// https://nextjs.org/docs/api-reference/next.config.js/introduction
 
 module.exports = {
   // Delegate static file compression to Nginx in production
-  compress: env.parsed['ENVIRONMENT'] === "prod" ? false : true,
-  env: env.parsed,
+  compress: process.env.ENVIRONMENT === "prod" ? false : true,
```
mattfredericksen, line 5: Do we want compression enabled in development? It seems like it would only be beneficial for network speeds, which are irrelevant for local dev containers.

frontend/pages/_app.tsx
```diff
@@ -1,27 +1,29 @@
...
     // Remove the server-side injected CSS.
-    // See https://github.com/mui-org/material-ui/blob/master/examples/nextjs/
+    // See https://github.com/mui-org/material-ui/blob/master/examples/nextjs/.
     const jssStyles = document.querySelector('#jss-server-side');
     if (jssStyles) {
       jssStyles.parentElement.removeChild(jssStyles);
     }
+    // TODO: get Django CSRF cookie?
+    // const url = "/api/csrf/set";
+    // axios.get(url).then(console.log);
```
mattfredericksen, line 17: Does this need a new issue in Jira / is the authentication functionally complete without it?

frontend/pages/_document.jsx
```diff
@@ -87,4 +91,4 @@ class MyDocument extends Document {
   }
 }
 
-export default MyDocument;
\ No newline at end of file
+export default MyDocument;
```
mattfredericksen, line 94: Since we [changed `MyApp` to `App`](https://github.com/ModularHistory/modularhistory/pull/289/files#diff-9c9ff648160c4ca56ad3e0809b4c493a3e9282c6546de0e4f7cf797ed8022560R7), should we change `MyDocument` to `Document`?

frontend/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,369 @@
...
+    cookies = cookies || null;
+  }
+  user.accessToken = accessToken;
+  user.refreshToken = refreshToken;
+  user.cookies = cookies;
+  return true;
+};
+
+// https://next-auth.js.org/configuration/callbacks#jwt-callback
+callbacks.jwt = async function jwt(token, user?: User, account?, profile?, isNewUser?: boolean) {
```
Comment (mattfredericksen) [inline frontend/pages/api/auth/[...nextauth].ts:245]: Do we not need types specified for account/profile? Would we want to add types for them eventually?

frontend/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,369 @@
...
+      }
+    });
+    token.accessTokenExpiry =
+      token.accessTokenExpiry ?? Date.parse(sessionTokenCookie.match(/expires=(.+?);/)[1]);
+  }
+  // Refresh the access token if it is expired.
+  if (Date.now() > token.accessTokenExpiry) {
+    token = await refreshAccessToken(token);
+  }
+  return Promise.resolve(token);
```
Comment (mattfredericksen) [inline frontend/pages/api/auth/[...nextauth].ts:268]: This is just a matter of syntax preference, but I think we could 
```javascript
return await token;
```

frontend/pages/api/auth/[...nextauth].ts
Comment (iacobfred) [reply frontend/pages/api/auth/[...nextauth].ts:245]: Probably so. I'll take a look when I enable social auth in MH-136.

frontend/pages/_document.jsx
iacobfred: Yup.

frontend/pages/_app.tsx
iacobfred: Ah. I'll create a Jira issue to circle back to it. Auth functionality is complete without it, BUT... Django only sends the CSRF cookie token when you visit a Django page. If we Reactify everything, it's possible that someone could visit the site and not receive a CSRF cookie from Django before it's required to make some POST request with DRF.

frontend/pages/auth/redirect.tsx
```diff
@@ -0,0 +1,30 @@
+import Container from '@material-ui/core/Container';
+import { useSession } from 'next-auth/client';
+import Head from 'next/head';
+import { useRouter } from 'next/router';
+import React from 'react';
+import Layout from '../../components/layout';
+
+const Redirect: React.FunctionComponent = () => {
+  const router = useRouter();
+  const [_session, loading] = useSession();
```
mattfredericksen, line 10: Why not just `session`?

frontend/pages/auth/signin.jsx
```diff
@@ -0,0 +1,93 @@
...
+      <Container>
+        <Box
+          display="flex"
+          justifyContent="center"
+          alignItems="center"
+          m={5}
+          p={5}
+          flexDirection="column"
+        >
+          {!loading && session && session.user && (
```
mattfredericksen, line 28: `{!loading && session?.user && (` ?

frontend/pages/entities/index.jsx
```diff
@@ -33,44 +32,52 @@ function EntitiesPagination({ count, ...childProps }) {
...
   const theme = useTheme();
 
-  const sibCount = 1 + ['sm', 'md'].map((size) =>
-    useMediaQuery(theme.breakpoints.up(size))
-  ).reduce((sum, current) => sum + current);
+  const sibCount =
+    1 +
+    ["sm", "md"]
+      .map((size) => useMediaQuery(theme.breakpoints.up(size)))
+      .reduce((sum, current) => sum + current);
```
mattfredericksen, line 39: ;-;

modularhistory/config/secrets.py
```diff
@@ -0,0 +1,4 @@
+from decouple import config
+
+# https://docs.djangoproject.com/en/3.1/ref/settings#s-secret-key
+SECRET_KEY = config('SECRET_KEY', default='f67EPexT9Tmwnt71kcGPk')
```
mattfredericksen, line 4: Your secret is now mine.

frontend/next.config.js
iacobfred: Hmm, I think I'd like for compression to be enabled one way or another in dev, if it's enabled one way or another in production. In production, it's handled by Nginx. In dev, I guess I'm fine with having it done by Next.js. (It's enabled by default, so the point here is just to disable it in prod where it's handled by Nginx.)

mattfredericksen (approved):

frontend/pages/auth/redirect.tsx
iacobfred: Just so the linter doesn't complain about it being unused.  : - )

PR #297, opened 2021-03-21
MH-7: Nginx cache for next.js / react upstream files

alashow:


alashow:
@iacobfred While doing this, I noticed a problem in `.config/nginx/prod/nginx.conf`:
https://github.com/ModularHistory/modularhistory/blob/dc74d056128779ce3483d3a02457a66a2b65c4a4/.config/nginx/prod/nginx.conf#L114-L118

I think, this `location` block actually being ignored because of other next.js/react upstream location block:
https://github.com/ModularHistory/modularhistory/blob/dc74d056128779ce3483d3a02457a66a2b65c4a4/.config/nginx/prod/nginx.conf#L131-L140

.config/nginx/dev/nginx.conf
```diff
@@ -25,6 +25,8 @@
 #     }
 # }
 
+proxy_cache_path /var/cache/nginx levels=1:2 keys_zone=STATIC_CACHE:10m inactive=7d use_temp_path=off max_size=10g;
```
iacobfred, line 28: @alashow, can you add a link to documentation for this directive?

.config/nginx/dev/nginx.conf
```diff
@@ -25,6 +25,9 @@
 #     }
 # }
 
+# http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_path
+proxy_cache_path /var/cache/nginx levels=1:2 keys_zone=STATIC_CACHE:10m inactive=7d use_temp_path=off max_size=10g;
```
iacobfred, line 29: @alashow, should we add a Docker volume to the `dev` service (in docker-compose.yml) so that the cache persists when containers are removed and recreated? (We would also have to figure out something for production environment, though, since production Nginx is in an LXC container.)

.config/nginx/dev/nginx.conf
```diff
@@ -68,14 +71,16 @@ server {
...
         proxy_pass http://react_server;
         proxy_redirect off;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto http;  # https in prod
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;
-        
+
         error_page 400 401 402 403 404 405 406 407 408 409 410 411 412 413 414 415 416 417 418 421 422 423 424 425 426 428 429 431 451 500 501 502 503 504 505 506 507 508 510 511 /error.htm;
+        add_header X-Status $upstream_cache_status; # dev only
```
iacobfred, line 83: @alashow, what does this do?

alashow:
Adds "HIT", "MISS" cache status headers to responses.

iacobfred:
> @iacobfred While doing this, I noticed a problem in `.config/nginx/prod/nginx.conf`:
> https://github.com/ModularHistory/modularhistory/blob/dc74d056128779ce3483d3a02457a66a2b65c4a4/.config/nginx/prod/nginx.conf#L114-L118
> 
> I think, this `location` block actually being ignored because of other next.js/react upstream location block:
> https://github.com/ModularHistory/modularhistory/blob/dc74d056128779ce3483d3a02457a66a2b65c4a4/.config/nginx/prod/nginx.conf#L131-L140

@alashow, are the `location` blocks processed in order?  Since the /_next/static/ rule comes first, shouldn't it be processed before the upstream location block?

.config/nginx/dev/nginx.conf
iacobfred: @alashow, maybe we could create another task for this in Jira.  (It could be lower priority than some other things you are working on.)

alashow:
> @alashow, are the `location` blocks processed in order? Since the /_next/static/ rule comes first, shouldn't it be processed before the upstream location block?

Apparently, it doesn't depend on the order of the blocks but it has a little complex selection algorithm that seems to depend on different things ([source](https://www.digitalocean.com/community/tutorials/understanding-nginx-server-and-location-block-selection-algorithms#matching-location-blocks), [source](https://stackoverflow.com/questions/30061661/nginx-location-block-selection-algorithm)). 

I tried testing with simple Nginx configurations involving locations blocks `location /_next/static/` and `location ~ ^/(_?_next.*|entities/?$)` and the latter always overrides the first one (order independent).

iacobfred:
@alashow, I noticed that this adds caching to the dev Nginx config but not to the production Nginx config; is that on purpose?

iacobfred:
> > @alashow, are the `location` blocks processed in order? Since the /_next/static/ rule comes first, shouldn't it be processed before the upstream location block?
> 
> Apparently, it doesn't depend on the order of the blocks but it has a selection algorithm that seems to depend on different things like the length of the match, whether it is regex/exact match etc. ([source](https://www.digitalocean.com/community/tutorials/understanding-nginx-server-and-location-block-selection-algorithms#matching-location-blocks), [source](https://stackoverflow.com/questions/30061661/nginx-location-block-selection-algorithm)).
> 
> I tried testing with simple Nginx configurations involving locations blocks `location /_next/static/` and `location ~ ^/(_?_next.*|entities/?$)` and the latter always overrides the first one (order independent).

Interesting!  cc @mattfredericksen

alashow:
> @alashow, I noticed that this adds caching to the dev Nginx config but not to the production Nginx config; is that on purpose?

Yeah, we need to fix `location /_next/static/` matching first because that's where we mainly needed the cache.

mattfredericksen:
@alashow @iacobfred I don't know if this has already been fixed (but not committed), but it looks like the simplest fix would be to modify https://github.com/ModularHistory/modularhistory/blob/dc74d056128779ce3483d3a02457a66a2b65c4a4/.config/nginx/prod/nginx.conf#L115 to `location ^~ /_next/static/`, which says don't look for any regex matches for this prefix.

iacobfred (approved):

PR #298, opened 2021-03-22
MH-128: Enabled debug-toolbar in dev env

alashow:
To enable: removed disqualifier for being dockerized and added static ip of dev service to internal ips

Also removed mongodb requirement for django service

### DIFF
--- a/modularhistory/config/_debug_toolbar.py
+++ b/modularhistory/config/_debug_toolbar.py
@@ -19,7 +19,7 @@ def show_toolbar(request: HttpRequest) -> bool:
         request.user.is_superuser,
     )
     disqualifiers = (
-        [settings.TESTING]
+        settings.TESTING,
     )
     if any(conditions) and not any(disqualifiers):
         return True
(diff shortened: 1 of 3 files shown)


modularhistory/config/_debug_toolbar.py
```diff
@@ -11,20 +11,15 @@
...
-        settings.DEBUG
-        and request.META.get('REMOTE_ADDR', None) in settings.INTERNAL_IPS,
+        settings.DEBUG and request.META.get('REMOTE_ADDR', None) in settings.INTERNAL_IPS,
         request.user.is_superuser,
     )
     disqualifiers = (
-        settings.TESTING,
-        # disable toolbar in Dockerized dev environment to avoid errors; TODO: fix?
-        settings.ENVIRONMENT == Environments.DEV and settings.DOCKERIZED,
+        [settings.TESTING]
```
iacobfred, line 22: I think this should be:
```
disqualifiers = (settings.TESTING,)
```
or
```
disqualifiers = (
    settings.TESTING,
)
```

alashow: I was getting `TypeError: 'bool' object is not iterable` errors, didn't know adding a comma `,` would fix that.

iacobfred: Ah, yes, the comma is how Python knows that it's a tuple even though the parentheses only contain one item.

iacobfred (approved):

PR #300, opened 2021-03-25
MH-26: Reactify the homepage and SERP

mattfredericksen:
https://modularhistory.atlassian.net/browse/MH-26


.config/nginx/dev/nginx.conf
```diff
@@ -67,7 +67,7 @@ server {
         auth_basic off;
     }
 
-    location ~ ^/(_?_next.*|entities/?$) {
+    location ~ ^/($|_?_next.*|entities/?$) {
```
iacobfred, line 70: @mattfredericksen, could you add a clarifying comment above this (indicating that it routes requests for the homepage and the entities page to the Next.js server)?

apps/occurrences/serializers.py
```diff
@@ -9,8 +9,8 @@ class OccurrenceSerializer(SearchableModelSerializer):
     """Serializer for occurrences."""
 
     date_html = serpy.Field()
-    summary = serpy.MethodField()
-    description = serpy.MethodField()
+    summary = serpy.Field(attr='summary.html')
```
iacobfred, line 12: This should be good, since `summary` and `description` are required fields on Occurrence.

alashow: Offtopic but how does `attr='summary.html'` exactly work? I [read](https://serpy.readthedocs.io/en/latest/api.html#serpy.Field) that `attr` is the attribute to get on the object but I couldn't understand where `summary.html` is processed (I'm guessing adding `.html` to field's must be doing HTML escaping or similar)

iacobfred: Yep, we have a custom HTMLField implementation (see https://github.com/ModularHistory/modularhistory/blob/4a187f44f12f33ab543a8fc2cb91d3236f5217d5/modularhistory/fields/html_field.py#L133). `summary` is an HTMLField on the Occurrence model, so `summary.html` returns the processed/escaped HTML for that field.

frontend/package.json
```diff
@@ -22,7 +22,6 @@
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "^4.14.0",
     "@typescript-eslint/parser": "^4.14.0",
-    "eslint": "^7.18.0",
```
iacobfred, line 25: @mattfredericksen ^

apps/search/api/views.py
```diff
@@ -1,8 +1,277 @@
...
+        super().__init__()
+        self.excluded_content_types = None
+        self.sort_by_relevance = False
+        self.suppress_unverified = False
+        self.entities = None
+        self.topics = None
+        self.places = None
+        self.results_count = 0
+
+    # def get_context_data(self, *args, **kwargs) -> Dict:
```
iacobfred, line 55: This can be deleted, right?

frontend/pages/_document.jsx
```diff
@@ -77,6 +77,9 @@ class MyDocument extends Document {
           {/*    </style>*/}
           {/*{% endif %}*/}
           {/*<script type="text/javascript" src='{% static "scripts/base.js" %}' defer></script>*/}
+          <style>
```
iacobfred, line 80: What does this do?  Can you add a comment (with a link to documentation if available)?

Can the commented-out code above this be deleted?

frontend/pages/entities/index.jsx
```diff
@@ -62,7 +63,7 @@ export default function Entities({ entitiesData }) {
     <Grid item key={entity['pk']}
       xs={6} sm={4} md={3}>
       <a href={`entities/${entity['pk']}`}>
-        <Card>
+        <Card elevation={3}>
```
iacobfred, line 66: I don't have a strong preference, but I wonder if we should make our own card component (extending MUI's component so that we can specify `elevation` (and anything else) in a single place.

mattfredericksen: I think we are very likely to do this eventually. As long as we're mostly using MUI, the best way to accomplish this will probably be with [MUI's theming](https://material-ui.com/customization/theming/) and [overriding default props](https://material-ui.com/customization/globals/).

apps/entities/api/urls.py
```diff
@@ -10,4 +10,5 @@
 
 urlpatterns = [
     path('', views.EntityListAPIView.as_view()),
+    path('partial/', views.EntityPartialAPIView.as_view()),
```
iacobfred, line 13: How about `EntityDetailAPIView` instead of `EntityPartialAPIView`?

.config/nginx/dev/nginx.conf
```diff
@@ -103,7 +103,13 @@ server {
...
     }
 
-    location ~ ^/(_?_next.*|entities/?$|api/auth/.*|auth/.*) {
+    # Routing regex for the NextJS server
+    #   `|` is an "or", separates patterns
+    #   `_?_next.*` matches any path starting with `_next` or `__next`
+    #   `entities/?$` matches entities page (localhost/entites/)
+    #      `?` is needed because the ending `/` is optional
+    # To add a route, insert `|your_path/?$` before the closing parenthesis
+    location ~ ^/($|_?_next.*|search/?$|entities/?$|api/auth/.*|auth/.*) {
```
mattfredericksen, line 112: This can be updated to route `/occurrences` and `/quotes` note.

apps/search/api/views.py
mattfredericksen: Sure it do

frontend/.eslintrc.js
```diff
@@ -16,7 +16,8 @@ module.exports = {
   plugins: ['react'],
   rules: {
     'no-unused-vars': ['warn', { varsIgnorePattern: '^_' }],
-    '@typescript-eslint/no-unused-vars': ['warn', { varsIgnorePattern: '^_' }]
+    '@typescript-eslint/no-unused-vars': ['warn', { varsIgnorePattern: '^_' }],
+    'react/react-in-jsx-scope': ['off'],
```
mattfredericksen, line 20: We don't need this rule since NextJS does magic or something so React is always in scope.

frontend/components/Footer.jsx
```diff
@@ -0,0 +1,66 @@
+import Link from "next/link";
```
mattfredericksen, line 1: This was just a file rename.

frontend/components/Navbar.jsx
```diff
@@ -18,28 +18,35 @@ const globalMenuItems = [
...
+    path: "/search/?content_types=occurrences.occurrence",
+    as: "/occurrences",
+    reactive: true
+  },
+  {
+    title: "Quotes",
+    path: "/search/?content_types=quotes.quote",
+    as: "/quotes",
+    reactive: true
+  },
```
mattfredericksen, line 32: Allows routing to `/occurrences` and `/quotes` without leaving the Next app using decorator urls (the `as` argument), though these routes do not exist yet, so refreshing routes to Django.

frontend/pages/entities/index.jsx
```diff
@@ -4,63 +4,11 @@ import CardHeader from "@material-ui/core/CardHeader";
...
-          {...item}
-          component={PaginationLink}
-          data-index={item.page}
-          href={item.page > 1 ? `/entities?page=${item.page}` : router.pathname}
-        />
-      )}
-      {...childProps}
-    />
-  );
-}
```
mattfredericksen, line 63: Moved to `frontend/components/Pagination.jsx`

frontend/.eslintrc.js
iacobfred: Oh, nice!

frontend/components/PageTransitionContext.jsx
```diff
@@ -0,0 +1,37 @@
+import {createContext, useEffect, useState} from "react";
+import {useRouter} from "next/router";
+
+// By placing `PageTransitionContext.Provider` at the root of
+// the application (pages/_app.tsx), any component can
+// use this context to react to page transitions without
```
iacobfred, line 6: `react` might be a confusing word to use here. : - )

frontend/components/search/SearchForm.jsx
```diff
@@ -0,0 +1,158 @@
...
+
+export default function SearchForm({contained}) {
+  const classes = useStyles({contained});
+  const router = useRouter();
+  const formState = useSearchFormState();
+
+  // This value is used to change how the form is displayed
+  // in a sidebar vs when loaded by itself.
+  // See: https://material-ui.com/components/grid/#grid-with-breakpoints
+  const sm = contained ? 12 : 6;
```
iacobfred, line 78: Why `contained`?  `contained` seems to imply that the component is being used inside a `Container`.  Is that the intent?

mattfredericksen: Updated comments & variable name for clarity

iacobfred (approved):

frontend/components/moduledetails/QuoteDetail.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+  let titleHtml = quote['attributee_html'];
+  if (quote['date_html']) titleHtml += `, ${quote['date_html']}`;
+
+  const firstImage = quote['serialized_images']?.[0];
+
+  return (
+    <>
+      <a className="edit-object-button float-right" target="_blank"
+         href="{% url 'admin:quotes_quote_change' quote.pk %}">
```
alashow, line 23: Rogue django-template code

frontend/components/search/CheckboxGroup.jsx
```diff
@@ -0,0 +1,60 @@
...
+  // `children` is an array of options, objects with keys `label` and `key`.
+  //    `label` is rendered next to the option.
+  //    `key` is the value of the option used in the API request.
+
+  // get current input values
+  const {state, setState, disabled} = useContext(SearchFormContext);
+
+  // If the current state is not set, default to all options.
+  // This is the default behavior expected by the API.
+  let checkedItems = state[name] || children.map(({key}) => key);
```
alashow, line 24: Not sure why, but this is currently not working on my end

frontend/components/search/SearchButton.jsx
```diff
@@ -18,14 +18,21 @@ export default function SearchButton(props) {
...
+        <CircularProgress size={12} style={{ marginRight: "10px" }} />
+        Loading
+      </span>
+    );
+  } else {
+    buttonText = "Search";
+  }
+
   return (
-    <Button variant={"contained"}
```
alashow: prettifier did this btw, not me)

frontend/components/moduledetails/OccurrenceDetail.jsx
```diff
@@ -0,0 +1,53 @@
...
+export default function OccurrenceDetail({occurrence}) {
+  return (
+    <>
+      {/*<a href="{% url 'admin:occurrences_occurrence_change' occurrence.pk %}"*/}
+      {/*   target="_blank" className="edit-object-button" rel="noopener noreferrer"*/}
+      {/*   style={{display: "inline-block", position: "absolute", top: "1px", right: "-2rem", fontWeight: "bold"}}>*/}
+      {/*  <i className="fa fa-edit"/>*/}
+      {/*</a>*/}
+
+      {occurrence['verified'] || (
```
alashow: This is probably nitpicking but is there a better way of doing this for easier reading? It was hard to understand at first that it's checking for the boolean field in `verified`

frontend/components/search/YearSelect.jsx
```diff
@@ -0,0 +1,48 @@
...
+  // `label` is displayed in the number input.
+  // `name` is the query parameter key used in the search API request.
+
+  const {state, setStateFromEvent, disabled} = useContext(SearchFormContext);
+
+  // `valueName` is the query param key for the year value (number).
+  // `typeName` is the query param key for the value type (e.g. "CE").
+  // These formats were chosen only because the API already supports them.
+  const valueName = `${name}_0`;
+  const typeName = `${name}_1`;
```
alashow, line 17: Replacing `_0` & `_1` suffixes with something like `_value` & `_type` would produce more readable/self-explanatory param keys.

frontend/pages/_document.jsx
```diff
@@ -52,13 +52,10 @@ class Document extends BaseDocument {
             href="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
           />
           {/* Material Design Bootstrap CSS */}
-          <link
```
alashow, line 55: Are we not using mdbootstrap anymore or was it moved somewhere else? In both cases, should we remove it from here?

alashow (changes requested):
.

frontend/components/moduledetails/OccurrenceDetail.jsx
mattfredericksen: Unfortunately, I think this is just a quirk of JSX syntax. This is the only way to do inline conditional rendering.
https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

The alternative would be do define a variable above:
```jsx
const unverifiedLabel = occurrence["verified"] ? undefined : <span> ... </span>;
```
And then to reference `{unverifiedLabel}` in the return JSX. But then it would not be immediately clear when looking at the return JSX that the label is conditionally rendered.

frontend/components/search/YearSelect.jsx
mattfredericksen: I agree. I'll create a Jira task for that.

frontend/components/moduledetails/OccurrenceDetail.jsx
mattfredericksen: https://modularhistory.atlassian.net/browse/MH-168?focusedCommentId=10240

apps/entities/api/views.py
```diff
@@ -17,5 +19,25 @@ class EntityViewSet(ModelViewSet):
...
+
+    pagination_class = VariableSizePagination
+    serializer_class = EntityPartialDictSerializer
+
+    def get_queryset(self):
+        attributes = self.request.query_params.getlist('attributes')
+        if not attributes:
+            return []
+
+        return Entity.objects.values(*attributes)
```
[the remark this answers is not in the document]
mattfredericksen: Ah, I noticed this but forgot about it. It seems unlikely that a user will manually type in a bad endpoint, and in development the error response is useful. I can add an `allowed_attributes` list that we can add to as necessary, and I'll see if I can set DRF to return a `400` response when parameters don't match it.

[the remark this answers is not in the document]
mattfredericksen: Fixed in 063721154a9d79c621c00614db9169b398e49588.

apps/search/api/views.py
```diff
@@ -1,8 +1,246 @@
...
+        Return the queryset/list of model instances.
+
+        Required because SearchResultsView inherits from the generic Django ListView.
+        """
+        request = self.request
+        self.sort_by_relevance = request.GET.get('ordering') == 'relevance'
+        self.suppress_unverified = request.GET.get('quality') == 'verified'
+
+        content_types = request.GET.getlist('content_types') or []
+        start_year = request.GET.get('start_year_0', None)
```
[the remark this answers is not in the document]
mattfredericksen: They were already being used in the Django SERP view, which is where I copied this code from. I didn't look for where these values were being used, but now that you have I may as well update them...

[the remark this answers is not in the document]
mattfredericksen: See 78ca763c6d7357be1706eaa0c7913e9c812d3d38

frontend/components/moduledetails/QuoteDetail.jsx
mattfredericksen: b8566406d6493cecfca58e58c85a007902d41e22

frontend/components/search/CheckboxGroup.jsx
mattfredericksen: 04696702d0cb18681ab54c35b34adae0d6f64828

frontend/components/search/YearSelect.jsx
mattfredericksen: 78ca763c6d7357be1706eaa0c7913e9c812d3d38

frontend/pages/_document.jsx
mattfredericksen: It is not being used anymore; it was overriding styles in Material-UI. I'd prefer to leave it commented out here for now, just to keep track of major changes between the React side and Django side. If we don't encounter any strange styling behavior, we can remove it soon.

iacobfred (approved):

alashow (approved):

PR #301, opened 2021-03-25
Added django-extensions and pygraphviz

alashow:
django-extensions has useful tools for debugging (like [`shell_plus`](https://django-extensions.readthedocs.io/en/latest/command_extensions.html)) and it's needed to [graph models](https://django-extensions.readthedocs.io/en/latest/graph_models.html).

To graph models, it's also required to install [Graphviz](https://pygraphviz.github.io/documentation/stable/install.html) on the host machine (`sudo apt-get install graphviz graphviz-dev
 ` or `brew install graphviz`).

To graph models:
`python3.7 manage.py graph_models -a -g -o graph.png`

Shell plus with SQL prints:
`python3.7 manage.py shell_plus --print-sql`


iacobfred:
> To graph models, it's also required to install [Graphviz](https://pygraphviz.github.io/documentation/stable/install.html) on the host machine (`sudo apt-get install graphviz graphviz-dev ` or `brew install graphviz`).

@alashow, in this case, should we also update `setup.sh` and `Dockerfile.django` to install Graphviz?  

I imagine that we could follow up on this by settings up a Celery task to automatically generate the model graph when models are updated.

alashow:
Thanks for adding installation steps to `setup.sh`. Integration tests are failing with error: 
```ruby
pygraphviz/graphviz_wrap.c:2711:10: fatal error: graphviz/cgraph.h: No such file or directory
       2711 | #include "graphviz/cgraph.h"
```

I think I had the same error why trying to install `pygraphviz` manually via poetry, which I solved by installing the `graphviz` itself via brew before running poetry.

iacobfred (approved):

iacobfred (approved):

PR #302, opened 2021-03-25
MH-130: Fix duplicate n queries on search form entities selections

alashow:
`id` and `type` was missing from searchable_fields array which was causing n queries

### DIFF
--- a/apps/entities/models/entity.py
+++ b/apps/entities/models/entity.py
@@ -91,7 +91,7 @@ class Entity(
         verbose_name_plural = 'Entities'
         ordering = ['name']
 
-    searchable_fields = ['name', 'aliases', 'description']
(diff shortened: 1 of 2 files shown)


apps/entities/models/entity.py
```diff
@@ -91,7 +91,7 @@ class Meta:
         verbose_name_plural = 'Entities'
         ordering = ['name']
 
-    searchable_fields = ['name', 'aliases', 'description']
+    searchable_fields = ['id', 'type', 'name', 'aliases', 'description']
```
iacobfred, line 94: @alashow, since `searchable_fields` is used to specify which fields should be checked in database queries, would it be better to add `id` and `type` elsewhere?  e.g., we could use `.only('id', 'type', *searchable_fields)`?

iacobfred (approved):

PR #315, opened 2021-04-10
MH-104: Added cachalot to cache sql queries

alashow:
Need more testing in production mode, but I don't expect any issues.

https://django-cachalot.readthedocs.io/en/latest/quickstart.html

Caches all queries made on tables except the ones blacklisted [here](https://github.com/ModularHistory/modularhistory/blob/182c5c0783206dfdf29676d90aaddd696bd82f04/modularhistory/config/caches.py#L44). It's not recommended to use it with frequently-updated tables. I didn't find tables like that so I tried to blacklist the tables that are not related to our codebase, like celery or django admin.


modularhistory/config/caches.py
```diff
@@ -10,24 +10,46 @@
 Cache = Dict[str, Any]
 CACHES: Dict[str, Cache]
 
+redisCache = {
```
iacobfred, line 13: nitpick:  Python convention is for constants to be in all caps -->
`REDIS_CACHE = {`

iacobfred: (And for other variable names, the JavaScript convention is camel case, but the Python convention is snake case.)

modularhistory/config/caches.py
```diff
@@ -10,24 +10,46 @@
...
+}
+
 # https://github.com/jazzband/django-redis
 if IS_DEV and use_dummy_cache:
     CACHES = {
         'default': {
             'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
-        }
+        },
+        'cachalot': REDIS_CACHE,
```
iacobfred, line 27: Was this just for testing?  Do we want to keep it?

pyproject.toml
```diff
@@ -150,6 +150,8 @@ Twisted = [
 uvicorn = { extras = ["standard"], version = "^0.12" }
 wordcloud = "^1.8.0"
 youtube-dl = "^2020.7.28"
+django-cachalot = "^2.3.5"
+Werkzeug = "^1.0.1"
```
iacobfred, line 154: What is Werkzeug for?

modularhistory/config/caches.py
alashow: I left it so it can be enabled easily in dev mode by changing `CACHALOT_CACHE`

pyproject.toml
alashow: it's for using `manage.py runserver_plus` (which can be configured to log queries via `--print_sql`)

iacobfred: Oh, I see; it's for runserver_plus, right?

modularhistory/config/caches.py
alashow: it should probably need to be improved to respect `use_dummy_cache`

pyproject.toml
iacobfred: Should we make this a dev-only dependency?

In production, we use Uvicorn.

alashow: yeah, I forgot to specify that when adding it.

iacobfred (approved):

iacobfred (approved):

iacobfred:
@alashow, to fix the conflict, you can locally rebase on main, then delete poetry.lock and regenerate it by running `poetry install`, then `git add poetry.lock`, then `git rebase --continue`.

PR #318, opened 2021-04-14
MH-173: refactoring & cleanup

iacobfred:


apps/sources/models/mixins/document.py
```diff
@@ -0,0 +1,54 @@
...
+        null=True,
+        blank=True,
+        help_text='URL for information regarding the document',
+    )
+
+    class Meta:
+        """Meta options for the _Engagement model."""
+
+        # https://docs.djangoproject.com/en/3.1/ref/models/options/#model-meta-options.
+        abstract = True
```
mattfredericksen, line 54: What is the purpose of this? I see it in [`PageNumberMixin`](https://github.com/ModularHistory/modularhistory/blob/a1bb56fadd7dd46fd139b09e984df0044fd1039c/apps/sources/models/mixins/page_numbers.py#L10) and [`TextualMixin`](https://github.com/ModularHistory/modularhistory/blob/a1bb56fadd7dd46fd139b09e984df0044fd1039c/apps/sources/models/mixins/textual.py#L8), but it doesn't look like either is related to [`_Engagement`](https://github.com/ModularHistory/modularhistory/blob/c4a89b375bc58c36eeefc3fccc2aa8753267ed2a/apps/entities/models/affiliation.py#L12).

mattfredericksen: Oh, or maybe the comment was just copied from somewhere else?

iacobfred: Oh, looks like that was copied and pasted. I'll update the docstrings.

`abstract = True` lets Django know that it shouldn't create a table for this model (i.e., it's an abstract model that should only be used for behavior inheritance with the Python objects).

mattfredericksen (approved):

PR #329, opened 2021-04-18
MH-176

aaw0151:
Converted Topics page to React


frontend/pages/topics/index.jsx
```diff
@@ -0,0 +1,49 @@
+import Container from "@material-ui/core/Container";
+import Grid from "@material-ui/core/Grid";
+import axios from "axios";
```
mattfredericksen, line 3: @iacobfred do we need to use [axiosWithoutAuth](https://github.com/ModularHistory/modularhistory/blob/main/frontend/axiosWithoutAuth.ts) here?

iacobfred: Hmm.  Looks like the entities page uses vanilla axios as well, and it works fine....

frontend/pages/topics/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+export default function Topics({ topicsData }) {
+  const topics = topicsData["results"] || [];
+
+  //Style for the anchor used in topicNames
+  const topicAnchorStyle = {
+    color: 'black',
+  }
+  const topicNames = topics.map((topic) => (
+    <Grid item key={topic['key']} xs={6} sm={6} md={6}>
```
mattfredericksen, line 15: Grid breakpoints apply to all larger sizes as well, so setting `xs={6}` also sets `sm`, `md`, and `lg` to `6`, unless those values are set explicitly. If instead these are set to `xs={6}`, `sm={4}`, and `md={3}`, then in larger viewports the grid will have more columns (the grid is always divided in 12 units).

frontend/pages/topics/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+      <a style={topicAnchorStyle} href={'./search?content_types=occurrences.occurrence&content_types=quotes.quote&content_types=images.image&content_types=sources.source&query=' + topic['key']}><u><strong>{topic['key']}</strong></u></a>
```
mattfredericksen, line 16: By default, all content types are selected in search results, so we should be able to drop the `content_types` parameters and get the same results. Also, rather than searching for the topic with the query string, we should use the `topics` paramater. E.g.: ``href={`/search?topics=${topic["pk"]}`}``. Finally, by setting this in a [`next/link`](https://nextjs.org/docs/api-reference/next/link) component we can utilize client-side routing instead of reloading the application. E.g.: 
```javascript
import Link from "next/link";
<Link href={`/search?topics=${topic["pk"]}`}>
  <a {...propsExceptHref}>...</a>
</Link>
```

frontend/pages/topics/index.jsx
aaw0151: xs has been set to 4 since I believe that 3 columns will look better. The other settings have been removed.

frontend/pages/topics/index.jsx
aaw0151: Queries have been updated and Link component has been added.

PR #330, opened 2021-04-18
fix auth bug

iacobfred:


core/settings.py
```diff
@@ -48,8 +48,6 @@
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 # https://docs.djangoproject.com/en/3.1/ref/settings#s-secure-ssl-redirect
 SECURE_SSL_REDIRECT = False  # SSL redirect is handled by Nginx reverse proxy in prod.
-# https://docs.djangoproject.com/en/dev/ref/settings/#session-cookie-httponly
-SESSION_COOKIE_HTTPONLY = IS_PROD
```
iacobfred, line 52: SESSION_COOKIE_HTTPONLY is True by default, and as experience has shown, it's best for us to have the same setting in dev environment as in prod (so that we aren't surprised by the different behavior after deploying).

frontend/pages/api/auth/[...nextauth].ts
```diff
@@ -166,12 +169,13 @@ callbacks.session = async function session(session: Session, jwt: JWT) {
 
 callbacks.redirect = async function redirect(url, baseUrl) {
   url = url.startsWith(baseUrl) ? url : baseUrl;
+  const reactPagesDoNotNeedToHitTheRedirectPage = false;
```
Comment (mattfredericksen) [inline frontend/pages/api/auth/[...nextauth].ts:172]: 😢

mattfredericksen (approved):

PR #331, opened 2021-04-18
MH-159

mattfredericksen:
https://modularhistory.atlassian.net/browse/MH-159


frontend/components/moduledetails/OccurrenceDetail.tsx
```diff
@@ -37,7 +44,9 @@ export default function OccurrenceDetail({ occurrence }) {
         />
         <div dangerouslySetInnerHTML={{ __html: occurrence["description"] }} />
 
-        {occurrence["postscript"] && <p dangerouslySetInnerHTML={occurrence["postscript"]} />}
```
mattfredericksen, line 40: This was a bug caught by TypeScript :)

frontend/tsconfig.json
```diff
@@ -23,7 +28,11 @@
     "*.ts",
     "*.tsx",
     "**/*.ts",
-    "**/*.tsx"
+    "**/*.tsx",
+    "*.js",
```
iacobfred, line 32: This will make our linter complain about Typescript violations in all the non-Typescript (js/jsx) files, but I think that's probably OK if it encourages us to finish converting those pages to Typescript. : - )

iacobfred (approved):

frontend/pages/[path].tsx
```diff
@@ -0,0 +1,53 @@
...
+interface StaticPageProps {
+  page: StaticPage;
+}
+
+const StaticPage: FC<StaticPageProps> = ({ page }: StaticPageProps) => {
+  // check if the viewport width is less than 600px
+  const isSmall = useMediaQuery("@media (max-width:600px)");
+
+  return (
+    <Layout>
```
Comment (iacobfred) [inline frontend/pages/[path].tsx:17]: @mattfredericksen, looks like this can't be compiled without the `title` attribute on Layout

Comment (mattfredericksen) [reply frontend/pages/[path].tsx]: Fixed in 4cbc0a7.

frontend/pages/occurrences/[slug].tsx
```diff
@@ -15,7 +15,7 @@ interface OccurrenceProps {
  */
 const Occurrence: FC<OccurrenceProps> = ({ occurrence }: OccurrenceProps) => {
   return (
-    <Layout title={""}>
+    <Layout title={occurrence["summary"] || occurrence["slug"]}>
```
Comment (iacobfred) [inline frontend/pages/occurrences/[slug].tsx:18]: I think `occurrence["summary"]` is probably sufficient, since summary is a required field, but @mattfredericksen, if you think it's a good idea to use slug as a fallback, should we go ahead and use a method to replace the slug's hyphens with spaces?

frontend/pages/quotes/[pk].tsx
```diff
@@ -15,7 +15,9 @@ interface QuoteProps {
  */
 const Quote: FC<QuoteProps> = ({ quote }: QuoteProps) => {
   return (
-    <Layout title={""}>
+    // TODO: add attribute to Quotes for title text (not html),
```
Comment (iacobfred) [inline frontend/pages/quotes/[pk].tsx:18]: @mattfredericksen, yeah... It'd be nice if we could query quotes by slug, and then the slug base could be used for this title attribute...
Maybe we could add a field named `key` (which could contain some user-added string identifier, e.g.,  "what is freedom to the slave") to the Quote model and update Quote's `get_slug` method to return the slugified key.

apps/quotes/serializers.py
```diff
@@ -37,7 +35,7 @@ def get_bite(self, instance: 'Quote') -> str:
         return (
             instance.bite.html
             if instance.bite
-            else truncatechars_html(instance.text, 100)
+            else ""
```
alashow, line 38: I hope this was an intentional change

frontend/pages/occurrences/[slug].tsx
```diff
@@ -0,0 +1,51 @@
...
+  let occurrence = {};
+  const { slug } = params;
+
+  await axios
+    .get(`http://django:8000/api/occurrences/${slug}/`)
+    .then((response) => {
+      occurrence = response.data;
+    })
+    .catch((error) => {
+      // TODO: how should we handle errors here?
```
Comment (alashow) [inline frontend/pages/occurrences/[slug].tsx:35]: What kind of errors do we expect here?

I guess it will return some kind of 'not found' error in case of invalid/nonexistent slugs. For 'not found' errors, we probably need to redirect to 404 pages on the client. We could generalize this for other endpoints with middleware maybe?

frontend/tsconfig.json
```diff
@@ -1,11 +1,16 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    // TODO: research if es2015 is supported by enough browsers
```
alashow, line 3: Reports https://caniuse.com/?search=es2015 suggest it's supported enough (for ex, chrome's last unsupported version's release was in 2015)

alashow (approved):

frontend/pages/occurrences/[slug].tsx
Comment (alashow) [reply frontend/pages/occurrences/[slug].tsx:18]: > Do you think we should truncate the summary here in case it's too long?

I think we should _not_ truncate it since it's a detail page (where else can someone see the full summary?), assuming the current design allows long summaries.

apps/quotes/serializers.py
mattfredericksen: :) yes; the truncated html of `instance.text` is already being set in the `truncated_html` field. I figured there's no need to return the same value twice, and this way we can detect whether a bite exists or not, and possibly use some logic depending on that.

frontend/pages/occurrences/[slug].tsx
Comment (mattfredericksen) [reply frontend/pages/occurrences/[slug].tsx:35]: Yeah I created [`MH-196`](https://modularhistory.atlassian.net/browse/MH-196) recently for creating [custom error pages](https://nextjs.org/docs/advanced-features/custom-error-page). I don't know what the best way to redirect to those pages is, though. My first thought would be to just pass the error along to the page component add a `useEffect` hook to redirect to the `404`/`500` page if an error exists, but this might not be the most efficient method.

frontend/tsconfig.json
mattfredericksen: Cool! I changed this to avoid some other complications with how TypeScript was compiling stuff, so if only very old browsers don't support it, I don't think we need to be concerned.

frontend/pages/occurrences/[slug].tsx
Comment (mattfredericksen) [reply frontend/pages/occurrences/[slug].tsx:18]: @alashow Ah, I mean should we truncate the summary that is inserted into the page title, not the one that is displayed on the page. After a quick google search, it seems like we want our page titles to stay under 60 characters. I think we should either intelligently truncate the title (i.e. on words, not characters), enforce a summary length limit in the model, or add a new attribute for page titles.

PR #332, opened 2021-04-19
MH-126: Donations and required components

sunzae99:
- Need a separate implementation of Braintree (currently running on Flask)
- The donate-now needs to route to a corresponding response page: either Success or Error.
- Needs css management (For now added to global.scss).
- Please let me know of any other errors that I can fix.


frontend/pages/donate-now/Error.jsx
```diff
@@ -0,0 +1,25 @@
+import Link from "next/link";
+import React from "react";
+import Layout from "../../components/Layout";
+
+class Error extends React.Component {
```
iacobfred, line 5: @sunzae99, for consistency with the rest of our frontend codebase, can you use function-based components?

It won't require changing much.  You can look at frontend/pages/entities/index.jsx for an example in JavaScript or (even better) frontend/pages/auth/redirect.tsx for an example in TypeScript.

iacobfred:
@sunzae99, to integrate with our Django backend (rather than Flask), we'll first need to create a donations "app."  You can do this as follows:

1. `docker-compose run django_helper python manage.py startapp donations` (This will create a new `donations` directory with some scaffolded files and subdirectories.)
2. `mv donations apps/donations` (This will move the donations directory into our `apps` directory.)
3. Open `apps/donations/apps.py`, and change the `name` attribute from `donations` to `apps.donations` (You can look at `apps/quotes/apps.py` as a reference.)
4. Append `'apps.donations.apps.DonationsConfig'` to the `INSTALLED_APPS` list in `core/settings.py`

Then, you can create an `api` directory in the donations app directory, and create a `urls.py` and `views.py` in the `api` directory. For an example, see the `apps/entities/api` directory.

After that, we can create API views (in `apps/donations/api/views.py`) to do what is currently being done by your Flask code.

mattfredericksen:
> @sunzae99, to integrate with our Django backend (rather than Flask), we'll first need to create a donations "app." You can do this as follows:
> 
> 1. `docker-compose run django_helper python manage.py startapp donations` (This will create a new `donations` directory with some scaffolded files and subdirectories. `python manage.py startapp <APP_NAME>` is the normal command, but we can use `docker-compose run django_helper` prefix to run it inside your django container.)
> 2. `mv donations apps/donations` (This will move the donations directory into our `apps` directory.)
> 3. Open `apps/donations/apps.py`, and change the `name` attribute from `donations` to `apps.donations` (You can look at `apps/quotes/apps.py` as a reference.)
> 4. Append `'apps.donations.apps.DonationsConfig'` to the `INSTALLED_APPS` list in `core/settings.py`
> 
> Then, you can create an `api` directory in the `apps/donations/` directory, and then you can create a `urls.py` and `views.py` in the `apps/donations/api/` directory. For an example, see the `apps/entities/api/` directory.
> 
> After that, we can create API views (in `apps/donations/api/views.py`) to do what is currently being done by your Flask code.

iacobfred:
@sunzae99, also, can you change the directory name of `donate-now` to `donations`?

I'm thinking we could use paths such as the following URL:

- /donate
- /donations/success
- /donations/error

PR #343, opened 2021-04-22
MH-166: Implement/test elasticsearch

alashow:


iacobfred (changes requested):

apps/search/documents.py
```diff
@@ -0,0 +1,125 @@
...
+DEFAULT_INDEX_SETTINGS = {'number_of_shards': 1, 'number_of_replicas': 0}
+
+
+@registry.register_document
+class QuoteDocument(Document):
+    class Index:
+        settings = DEFAULT_INDEX_SETTINGS
+        name = 'quotes'
+
+    text = fields.TextField(attr='text.text', analyzer=html_strip)
```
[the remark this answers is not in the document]
alashow: Yes. For some reason, I thought `text` was the raw value.

apps/search/documents.py
```diff
@@ -0,0 +1,125 @@
...
+        'aliases': fields.TextField(),
+        'description': fields.TextField(attr='description.text', analyzer=html_strip)
+    })
+
+    class Django:
+        model = Occurrence
+        related_models = [Source, Entity]
+
+    def get_queryset(self):
+        return super(OccurrenceDocument, self).get_queryset().prefetch_related(
```
iacobfred, line 82: nitpick: We can use `super().get_queryset()` rather than `super(OccurrenceDocument, self).get_queryset()`

apps/occurrences/manager.py
```diff
@@ -1,6 +1,7 @@
 from typing import List, Optional
 
 from django.db.models import Prefetch, Q
+from elasticsearch_dsl import Q as Q2
```
iacobfred, line 4: How about `as ESQuery` for readability?

alashow: We won't need the other `Q` in there after ES implementation

apps/occurrences/manager.py
```diff
@@ -21,13 +22,16 @@ def search(
         suppress_hidden: bool = True,
     ) -> 'SearchableModelQuerySet':
         """Return search results from apps.occurrences."""
+
+        if not query:
+            es_qs = super().search(suppress_unverified=suppress_unverified, suppress_hidden=suppress_hidden)
```
iacobfred, line 27: If the queryset doesn't necessarily comes from ElasticSearch, can we rename `es_qs` to just `qs`?

docker-compose.yml
```diff
@@ -245,6 +245,39 @@ services:
     volumes:
       - "redisinsight:/db"
 
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.8.1
+    environment:
+      - node.name=modularhistory-es
```
iacobfred, line 251: @alashow, where do these settings come from?  Can you add a comment with a link to documentation, if available?

docker-compose.yml
```diff
@@ -256,4 +289,5 @@ volumes:
   # `data` is used by both redis and mongodb
   data:
   postgres_data:
+  search-data:
```
iacobfred, line 292: Can we make it `search_data`?

docker-compose.yml
```diff
@@ -245,6 +245,39 @@ services:
     volumes:
       - "redisinsight:/db"
 
+  elasticsearch:
```
iacobfred, line 248: @alashow, can you add `elasticsearch` as a dependency of the `django` service?

apps/search/documents/config.py
```diff
@@ -0,0 +1,10 @@
+from elasticsearch_dsl import analyzer
+
+html_strip = analyzer(
```
iacobfred, line 3: For clarity, could we rename this to something like `html_field_analyzer`?

core/constants/content_types.py
```diff
@@ -47,7 +47,7 @@ class ContentTypes(Constant):
 }
 
 
-def get_ct_id(ct_name: str) -> int:
+def get_ct_id(ct_name: str) -> ContentType:
```
iacobfred, line 50: Nice catch @alashow. Could you actually update the `ct_id = ...` line, so that it gets the id rather than the ContentType instance)?

i.e., `ct_id = ContentType.objects.get_by_natural_key(app_label=app_label, model=model).id`

alashow: If the current version works, should we even change it? The function is mostly used to pass ids in queryset filters, and it seems to be the filters can correctly use given models (I'm guessing this is normal django models behavior?).

iacobfred: Yeah, Django will always convert `<FK_FIELD_NAME>` to `<FK_FIELD_NAME>_id` when making queries, and it can also read the `id` of a model instance that is assigned to `<FK_FIELD_NAME>_id`, which is probably why it works currently. I think I read somewhere that it's better to use the `id` than the full model instance, though, so that you don't have to keep the model instance in memory or something... I'm having trouble finding it in the Django docs, though.

apps/search/api/filters/post_resolve_filters.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Filter that sorts queryset by SORT_BY_PARAM
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        sort_by_date = request.query_params.get(SORT_BY_PARAM) == 'date'
+
+        key = date_sorter if sort_by_date else score_sorter
+
+        return sorted(queryset, key=key, reverse=not sort_by_date)
```
iacobfred, line 50: @alashow, the sorting has to be reverse if it's based on relevance score?

apps/search/api/filters/post_resolve_filters.py
```diff
@@ -0,0 +1,73 @@
...
+def date_sorter(model_instance: Union[SearchableDatedModel, Dict]) -> HistoricDateTime:
+    """Return the value used to sort the model instance by date."""
+    get_date = getattr(model_instance, 'get_date', None)
+    if get_date is not None:
+        date = get_date()
+    elif isinstance(model_instance, dict):
+        date = model_instance.get('date')
+    else:
+        date = getattr(model_instance, 'date', None)
+    if not date:
```
iacobfred, line 62: Maybe we should log an error here, since we do expect for `model_instance` to have `date` or `get_date`.

How about this:
```
logging.error(
    f'{model_instance} has no date attribute '
    'but is included in search results.'
)
```

apps/search/api/filters/pre_resolve_filters.py
```diff
@@ -0,0 +1,31 @@
+
+SORT_BY_PARAM = 'ordering'
+
+# list of filters that's applied to every module
+PRE_RESOLVE_FILTERS = [
+    # random ordering filter could be applied as a hack to disable caching search queries via cachalot
```
iacobfred, line 6: @alashow, do you think it's actually a problem that cachalot caches search queries? Also, can you explain what is meant by "random queries"?

alashow: Depends on server memory + search usage. We're not actually gaining anything from caching search results because they're already fast enough. It will only be a problem when there's a lot of search queries.

apps/search/api/filters/post_resolve_filters.py
alashow: Yes. If `reverse=True` is ascending order:
1. date sort sorts from beginning time to end time = `asc`
2. relevance score sorts from highest score to lowest score = `descending`

apps/search/api/search.py
```diff
@@ -0,0 +1,77 @@
...
+        for result in s:
+            index = result.meta.index
+            result_groups.setdefault(index, []).append(result)
+            key = f"{index}_{result.meta.id}"
+            self.results_by_id[key] = result
+
+        # build queryset chain for each result group by resolving es results to django models
+        qs = chain()
+        for index, result_group in result_groups.items():
+            if index not in SEARCHABLE_DOCUMENTS:
```
iacobfred, line 56: This could be refactored to this:
```
document = SEARCHABLE_DOCUMENTS.get(index)
if not document:
    logging.error(
        f"Couldn't find document definition for this index = {index}"
    )
    continue
```

iacobfred:
@alashow, I added a new task, `update-git-hooks`, so that you can update your git hooks without running the setup script.

In this branch, can you run `git pull` and then `poetry run setup.update-git-hooks`?

apps/search/api/search.py
```diff
@@ -0,0 +1,77 @@
...
+
+class Search(DSLSearch):
+    results_count: int
+    results_by_id: Optional[dict]
+
+    def to_queryset(self, view):
+        """
+        Resolves elasticsearch results to django models.
+        """
+        s = self
```
iacobfred, line 30: Would it be OK to just use `self` for readability, or is this necessary to keep original attribute values for `self` or something?

apps/search/api/serializers.py
```diff
@@ -12,3 +16,15 @@ class SearchableModelSerializer(ModelSerializer):
...
     verified = serpy.BoolField()
+    meta = serpy.MethodField()
+
+    def get_meta(self, model) -> Optional[dict]:
+        if not hasattr(model, 'meta'):
+            return None
+
+        meta = model.meta.to_dict()
+        for key in ELASTICSEARCH_META_FIELDS_TO_CLEAN:
+            # TODO: find out why in some cases it's already cleaned
```
iacobfred, line 27: If you want, we could use a try-except:
```
try:
    del meta[key]
except KeyError as err:
    logging.info(
        f'The "{key}" key has already been removed from model meta.'
    )
```

apps/search/documents/base.py
```diff
@@ -0,0 +1,12 @@
...
+from django_elasticsearch_dsl import fields
+
+
+class Document(ESDocument):
+
+    hidden = fields.BooleanField()
+    verified = fields.BooleanField()
+
+    @classmethod
+    def index_name(cls, index=None):
```
iacobfred, line 11: @alashow, can you name this `get_index_name`? 
(I like the convention of `thing.property` and `thing.get_property()`)

apps/search/api/search.py
alashow: We can't just use self because of `s = s.execute()` line, where we need save the response into a variable. I renamed `s` to `response` for better readability (commit pending)

apps/search/api/serializers.py
alashow: I would like to keep logs clean for now because these always occur.

iacobfred (approved):

PR #345, opened 2021-04-23
MH-170: implement GraphQL

aaw0151:


core/settings.py
```diff
@@ -299,6 +301,9 @@
 MEGA_DEV_USERNAME = config('MEGA_DEV_USERNAME', default=MEGA_USERNAME)
 MEGA_DEV_PASSWORD = config('MEGA_DEV_PASSWORD', default=MEGA_PASSWORD)
 
+# GraphQL
+GRAPHENE = {'SCHEMA': 'apps.topics.models.schema.schema'}
```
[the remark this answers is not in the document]
iacobfred: Nm, it's correct!

iacobfred (changes requested):

apps/topics/models/schema.py
```diff
@@ -0,0 +1,40 @@
...
+    all_topics = graphene.List(TopicType)
+
+    def resolve_all_topics(self, info, **kwargs):
+        return Topic.objects.all()
+
+
+schema = graphene.Schema(query=Query)
+
+
+def query_result():
```
mattfredericksen, line 30: Is this function only for testing purposes? My understanding is that we would want all request schemas defined on the frontend, so that control over what data is retrieved is always held by the frontend. Is that correct?

iacobfred: Good point @mattfredericksen.

@aaw0151, I just looked at a couple tutorials, and it looks like most developers use Apollo to interact with GraphQL APIs from a React page.

This is the best walkthrough I found: https://zoejoyuliao.medium.com/django-graphql-react-2-integrate-graphql-into-your-react-project-71fa74f1cb00

apps/topics/api/urls.py
```diff
@@ -9,6 +9,6 @@
 app_name = 'topics'
 
 urlpatterns = [
-    path('', views.TopicListAPIView.as_view()),
+    path('', views.TopicGraphQLView.as_view()),
```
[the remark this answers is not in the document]
aaw0151: Resolved in latest commit.

frontend/pages/topics/index.jsx
```diff
@@ -37,7 +37,7 @@ export async function getServerSideProps(context) {
   let topicsData = {};
 
   await axios
-    .get("http://django:8000/api/topics/", { params: context.query })
+    .get("http://django:8000/graphql/?query={allTopics{key%20pk}}", { })
```
iacobfred, line 40: Does axios require commas to be replaced with %20?

apps/topics/schema.py
```diff
@@ -0,0 +1,10 @@
+import graphene
+
+import apps.topics.models.schema as topics_schema
```
iacobfred, line 3: nitpick: I'm not exactly sure why, but I've read that `from ... import ...` syntax is preferred. In this case: 
`from apps.topics.models import schema as topics_schema`

apps/topics/models/schema.py
aaw0151: The schema will be updated in the next implementation to a more "generic" schema to allow for easier additions.

apps/topics/models/schema.py
```diff
@@ -0,0 +1,40 @@
...
+
+# GraphQL query for all Topics with the fields of 'key' and 'pk'
+class Query(graphene.ObjectType):
+    all_topics = graphene.List(TopicType)
+
+    def resolve_all_topics(self, info, **kwargs):
+        return Topic.objects.all()
+
+
+topicsSchema = graphene.Schema(query=Query)
```
iacobfred, line 27: nitpick: While JavaScript prefers camel case (`topicsSchema`), Python prefers snail case (`topics_schema`)

apps/topics/models/schema.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def query_result():
```
iacobfred, line 30: @aaw0151, can you delete this method from schema.py, since we are building the query in the React page instead?

iacobfred (changes requested):
- https://github.com/ModularHistory/modularhistory/pull/345/files#r619604870
- https://github.com/ModularHistory/modularhistory/pull/345/files#r619605002
- https://github.com/ModularHistory/modularhistory/pull/345/files#r619604066

iacobfred (approved):

PR #361, opened 2021-05-01
Much!

iacobfred:


apps/images/api/schema.py
```diff
@@ -0,0 +1,31 @@
...
+from apps.images.models import Image
+
+
+class Query(graphene.ObjectType):
+    """GraphQL query for all images."""
+
+    images = graphene.List(ImageType)
+    image = graphene.Field(ImageType, slug=graphene.String())
+
+    def resolve_images(self, info, **kwargs):
```
mattfredericksen, line 15: Should we use `parent` instead of `self` on our `resolve_*` methods?
https://docs.graphene-python.org/en/latest/types/objecttypes/#implicit-staticmethod

apps/postulations/api/views.py
```diff
@@ -1,5 +1,5 @@
 from rest_framework import permissions
-from rest_framework.generics import ListAPIView
+from rest_framework.generics import ListAPIView, RetrieveAPIView
```
mattfredericksen, line 2: Is `RetrieveAPIView` needed?

apps/images/api/schema.py
iacobfred: It looks like in the graphene-django docs (e.g., https://docs.graphene-python.org/projects/django/en/latest/queries/), they sometimes name the parameter `root` and sometimes `self`. It won't affect the behavior, but do you think it would be more readable with `root` (and/or the `@staticmethod` decorator)?

apps/postulations/api/views.py
iacobfred: Nope.

apps/quotes/api/views.py
```diff
@@ -24,4 +24,5 @@ class QuoteAPIView(RetrieveAPIView):
     """API view for a single occurrences."""
 
     queryset = Quote.objects.all()
+    lookup_field = 'slug'
```
mattfredericksen, line 27: Do we want this specified? IIRC it is inferred from the route parameter by default.

iacobfred: IIRC, it didn't work with the `slug` field until I specified `lookup_field` (after searching on StackOverflow)... But I think we'll get rid of this view soon anyway (to use GraphQL consistently).

apps/sources/api/types.py
```diff
@@ -0,0 +1,19 @@
...
+
+class SourceType(AbstractModuleType):
+    """GraphQL type for the Source model."""
+
+    slug = graphene.String()
+    citation_html = graphene.String()
+    citation_string = graphene.String()
+    attributee_html = graphene.String()
+    title = graphene.String()
+    model = graphene.String()
```
mattfredericksen, line 15: Isn't this inherited? https://github.com/ModularHistory/modularhistory/blob/6ee5a052b250c0e24893fd635196d8acde90d026/apps/graph/types.py#L5-L8

iacobfred: Ah, yes, that's a relic from immediately before I implemented AbstractModuleType.

frontend/components/PageHeader.tsx
```diff
@@ -0,0 +1,16 @@
...
+const PageHeader: FC<PageHeaderProps> = ({ children }: PageHeaderProps) => {
+  return (
+    <>
+      <h1 className="page-header text-center my-3">{children}</h1>
+      <hr />
+    </>
+  );
+};
+
+export default PageHeader;
```
mattfredericksen, line 16: Could we replace the header on the SERP with this component? They look a bit different.
https://github.com/ModularHistory/modularhistory/blob/6ee5a052b250c0e24893fd635196d8acde90d026/frontend/pages/search.jsx#L74-L84
image

apps/images/api/schema.py
mattfredericksen: I don't know enough atm to have an opinion about whether `root` or `parent` would be better. I don't like using `@staticmethod` if it does not change behavior, but it would make it clear that we don't have access to `self`. Mostly, I just don't like the variable being named `self` if it is not `self`.

frontend/components/PageHeader.tsx
iacobfred: I do think it would be good to update the SERP header. This new component isn't a good drop-in replacement for what's being use now, though, so we should probably make a separate task for it.

apps/occurrences/serializers.py
```diff
@@ -8,13 +8,13 @@
...
+    dateHtml = serpy.Field(attr='date_html')
     summary = serpy.Field(attr='summary.html')
     description = serpy.Field(attr='description.html')
     postscript = serpy.MethodField()
-    serialized_images = serpy.Field()
-    primary_image = serpy.Field()
-    serialized_citations = serpy.Field()
+    serializedImages = serpy.Field(attr='serialized_images')
+    primaryImage = serpy.Field(attr='primary_image')
+    serializedCitations = serpy.Field(attr='serialized_citations')
```
mattfredericksen, line 17: Is it safe to update the case on these as well?
https://github.com/ModularHistory/modularhistory/blob/da66384674d97e0ea2f436b57a313b1ed6d7d184/apps/search/models/searchable_model.py#L65-L73

iacobfred: Yep!

apps/quotes/serializers.py
```diff
@@ -16,28 +16,23 @@ class QuoteSerializer(SearchableModelSerializer):
 
     bite = serpy.MethodField()
     html = serpy.Field()
-    truncated_html = serpy.MethodField()
     has_multiple_attributees = serpy.BoolField()
```
mattfredericksen, line 19: Do we not want `hasMultipleAttributees`?

frontend/components/cards/ModuleCard.tsx
```diff
@@ -0,0 +1,180 @@
+import { BaseModule } from "@/interfaces";
+import { Card } from "@material-ui/core";
+import { makeStyles } from "@material-ui/styles";
+import { FC, ReactNode } from "react";
+
+const useStyles = makeStyles({
```
mattfredericksen, line 6: Created [`MH-212`](https://modularhistory.atlassian.net/browse/MH-212) for moving this styling to a CSS module.

apps/quotes/serializers.py
iacobfred: Looks like it's not being used in any front-end code.

frontend/components/cards/ModuleUnionCard.jsx
```diff
@@ -0,0 +1,42 @@
...
+      return <ImageCard image={module} {...childProps} />;
+    case "occurrences.occurrence":
+      content = <div dangerouslySetInnerHTML={{ __html: module["summary"] }} />;
+      break;
+    case "postulations.postulation":
+      content = <div dangerouslySetInnerHTML={{ __html: module["summary"] }} />;
+      break;
+    case "quotes.quote":
+      content = (
+        <>
```
mattfredericksen, line 20: Looks like this fragment can be removed.

frontend/components/details/ImageDetail.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+interface ImageDetailProps {
+  image: ImageModule;
+}
+
+const ImageDetail: FC<ImageDetailProps> = ({ image }: ImageDetailProps) => {
+  const firstImage = image["serializedImages"]?.[0];
+  return (
+    <>
+      <h1 className="text-center card-title" dangerouslySetInnerHTML={{ __html: image["name"] }} />
```
mattfredericksen, line 13: Do images have a `name`? I can't find where that's coming from.
https://github.com/ModularHistory/modularhistory/blob/5b6ae545f3c62904f58d71f7e7b37e7636363324/apps/images/models/image.py#L51
https://github.com/ModularHistory/modularhistory/blob/6f102cd1e165f2a00240a649b0e6b779244a9507/apps/images/api/types.py#L7

frontend/components/details/ModuleDetail.tsx
```diff
@@ -1,13 +1,30 @@
...
 
 interface ModuleDetailProps {
-  module: OccurrenceModule | QuoteModule | EntityModule | TopicModule;
+  module:
+    | OccurrenceModule
+    | QuoteModule
+    | EntityModule
+    | TopicModule
+    | PostulationModule
+    | ImageModule;
```
mattfredericksen, line 27: We'll probably want to add
```ts
export type ModuleUnion = ... | ... | ...
```
to `interfaces.d.ts` when `ModuleUnionCard` gets converted to TS.

frontend/components/cards/ModuleCard.tsx
```diff
@@ -0,0 +1,180 @@
...
+      },
+    },
+  },
+});
+
+interface ModuleCardProps {
+  module: BaseModule;
+  header?: string;
+  className?: string;
+  style?: Record<string, any>;
```
mattfredericksen, line 128: This should be `CSSProperties` (import from `react`).

frontend/components/details/ImageDetail.tsx
iacobfred: Fixed.

frontend/pages/postulations/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+};
+
+export default Postulations;
+
+// https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering
+export const getServerSideProps: GetServerSideProps = async (context) => {
+  let postulationsData = {};
+
+  await axiosWithoutAuth
+    .get("http://django:8000/api/postulations/", { params: context.query })
```
mattfredericksen, line 51: Does this need to be changed to graphql? We're referencing attributes above (e.g. `slug`) that are not in the `PostulationSerializer`.
https://github.com/ModularHistory/modularhistory/blob/01a44f9e0be0a64d8bb76c7ff4284d7d66f9a2ac/apps/postulations/serializers.py#L8-L19

iacobfred: We can't use GraphQL for list pages yet, since our GraphQL API doesn't return the `total_pages` field used by the pagination component, so for now, I'll just update the serializer.

mattfredericksen (approved):

PR #382, opened 2021-05-09
MH-203: Refactoring the basic donations page, needs conditional rendering

sunzae99:
Made changes to the styling of the pages, but still need conditional rendering fixed. I have tried the code and is commented out. I was not entirely sure why the code was not working. I will really appreciate anyone's input.


frontend/pages/donations/make.tsx
```diff
@@ -1,15 +1,30 @@
...
+    marginBottom: "2rem",
+  },
+});
 interface DonateProps {
   clientToken: string;
 }
 
 const Donate: FC<DonateProps> = (props: DonateProps) => {
+  const classes = useStyles();
   const [clientToken, setClientToken] = useState(props.clientToken);
```
iacobfred, line 28: @sunzae99, maybe we should use state for `success` and `error`, to help with conditional rendering.

We could do it like this:
```
const [success, setSuccess] = useState(false); 
const [error, setError] = useState(false); 
```

frontend/pages/donations/make.tsx
```diff
@@ -45,7 +91,9 @@ const Donate: FC<DonateProps> = (props: DonateProps) => {
         });
         if (response.data == "OK") {
           router.push("/donations/success");
+          // return <SuccessMessage />;
```
iacobfred, line 94: @sunzae99, if we use state (suggested in https://github.com/ModularHistory/modularhistory/pull/382/files#r629693380), we could do this:
```
if (response.data == "OK") {
    setSuccess(true);
} else {
    setError(true);
}
```

mattfredericksen (changes requested):
Before we can allow users to navigate to this page in production, we'll need some more input validation (e.g. don't allow API requests with no amount entered, or a negative amount entered), but we probably don't need that on this PR.

frontend/pages/donations/make.tsx
```diff
@@ -12,7 +12,12 @@ interface DonateProps {
 const Donate: FC<DonateProps> = (props: DonateProps) => {
```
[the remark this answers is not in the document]
mattfredericksen: That is not completely true actually. We really just want to clearly indication which component is the core page component, so I would suggest naming this component `MakeDonationPage` or something similar with `Page` on the end.

mattfredericksen (approved):
LGTM!

iacobfred (approved):
LGTM! We'll want to make some changes to the text of the error message and success message, but we can do that in a future PR.

sunzae99:
> LGTM! We'll want to make some changes to the text of the error message and success message, but we can do that in a future PR.

What changes would you like to see? @iacobfred @mattfredericksen. I am free to work with your guys' suggestions :)

iacobfred:
> > LGTM! We'll want to make some changes to the text of the error message and success message, but we can do that in a future PR.
> 
> What changes would you like to see? @iacobfred @mattfredericksen. I am free to work with your guys' suggestions :)

I'd like to think more about the textual content of the donations page, but we don't need to worry about that right now. Good work @sunzae99. Let's merge it. : - )

PR #559, opened 2021-07-27
Fix MUI TextField style

mattfredericksen:
Fixes an issue where the top outline of a `TextField` would be erased by bootstrap styles on the `legend` html element.
For some reason I also converted the Navbar component to TSX.

Before:
image
image

After:
image
image


frontend/components/Navbar.tsx
```diff
@@ -45,37 +57,40 @@ function WrappedNavLink({ title, path, as, reactive }) {
...
+  menuItems?: GlobalMenuItem[];
+}
+
+const GlobalNavbar: FC<GlobalNavbarProps> = ({ menuItems }: GlobalNavbarProps) => {
   menuItems = menuItems || globalMenuItems;
 
   const router = useRouter();
-  const [session, loading] = useSession();
+  // TODO: Create session type and remove this cast to `any`.
+  const [session, loading] = useSession() as any;
```
mattfredericksen, line 93: Casting for now, since otherwise we get complaints when trying read attributes like `session.user["avatar"]`.

frontend/styles/globals.css
```diff
@@ -17,22 +17,25 @@ body {
...
+}
+
 #footer a {
   color: white;
 }
 
-a:hover {
-  text-decoration: none;
+legend {
+  width: revert;
```
mattfredericksen, line 36: This is what fixes the style problem. Since we import this file after importing bootstrap CSS, this resets to browser defaults. The `revert` is correctly overridden by other component-level styles.

iacobfred: Should we add a comment here?

iacobfred (approved):

iacobfred (approved):

frontend/components/Navbar.tsx
```diff
@@ -2,19 +2,27 @@ import { Divider } from "@material-ui/core";
...
 
-const globalMenuItems = [
+interface GlobalMenuItem {
+  title: string;
+  path: string;
+  reactive?: boolean;
+  submenuItems?: GlobalMenuItemWithoutSubmenuItems[];
+}
+
+type GlobalMenuItemWithoutSubmenuItems = Omit<GlobalMenuItem, "submenuItems">;
```
mattfredericksen, line 19: This is verbose (👎) but clear and removes the need to repeat the Omit in a few places (👍).

frontend/components/Navbar.tsx
```diff
@@ -34,7 +42,17 @@ const globalMenuItems = [
   },
 ];
 
-function WrappedNavLink({ title, path, as, reactive }) {
+function hasSubmenuItems(
+  menuItem: GlobalMenuItem
+): menuItem is GlobalMenuItem & Required<Pick<GlobalMenuItem, "submenuItems">> {
+  return menuItem.submenuItems !== undefined;
+}
```
mattfredericksen, line 49: We need a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) here since TS cannot guarantee that the properties of an object have not been modified after checking their type.

iacobfred (approved):

iacobfred (approved):

frontend/components/Navbar.tsx
iacobfred: alrighty

PR #564, opened 2021-07-29
Remove `page` from search form state

mattfredericksen:
This was causing the page number to not be reset when search terms were modified.


iacobfred (approved):

iacobfred:
amazing

PR #568, opened 2021-08-02
Content moderation backbone

iacobfred:


apps/topics/models/taggable_model.py
```diff
@@ -56,14 +54,14 @@ def tags_string(self) -> str:
...
-    def tags_html(self) -> SafeString:
+    def tags_html(self) -> str:
         """Return the model instance's tags as an HTML string of <li> elements."""
         tags_html = ''
         if self.tag_keys:
             tags_html = ' '.join(
                 [f'<li class="topic-tag"><a>{tag_key}</a></li>' for tag_key in self.tag_keys]
             )
-        return format_html(tags_html)
+        return tags_html
```
alashow, line 64: Is formatting not needed anymore or maybe done somewhere else now?

alashow (commented):
LGTM! I didn't try deploying it yet, but the code logic about changes/moderations seemed okay.

apps/topics/models/taggable_model.py
iacobfred: It's not needed anymore, since we are no longer using Django templates for pages that display topic tags. (`format_html` is just a helper for escaping HTML that's rendered in Django templates.)

apps/donations/api/views.py
```diff
@@ -7,9 +7,7 @@
     braintree.Environment.Sandbox,
     merchant_id=config('BRAINTREE_MERCHANT_ID', default='hqscdtx5y45q8gst'),
     public_key=config('BRAINTREE_PUBLIC_KEY', default='gqwntn8ffm8g6pcf'),
-    private_key=config(
-        'BRAINTREE_PRIVATE_KEY', default='52bd7b66686e3d302208ca55ddac87c2'
-    ),
+    private_key=config('BRAINTREE_PRIVATE_KEY', default='52bd7b66686e3d302208ca55ddac87c2'),
```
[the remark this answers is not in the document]
iacobfred: Hmm, they might be the keys for our sandbox account... I'll check now. If they're our sandbox keys, we should still take them out of source code, though; I'll add a commit.

apps/moderation/diff.py
```diff
@@ -0,0 +1,138 @@
...
+        return render_to_string(template, context)
+
+
+class TextChange(FieldChange):
+    """A change to the value of a text field."""
+
+    @property
+    def diff(self) -> str:
+        """Render the diff."""
+        value_before, value_after = escape(self.change[0]), escape(self.change[1])
```
mattfredericksen, line 37: map(escape, self.change) maybe?

apps/dates/fields.py
```diff
@@ -17,65 +16,57 @@ class HistoricDateTimeField(DateTimeField):
...
-                historic_datetime.day,
-                historic_datetime.hour,
-                historic_datetime.minute,
-                historic_datetime.second,
-                historic_datetime.microsecond,
-            )
+        elif isinstance(value, str):
+            value = isoparse(value)
+        if isinstance(value, HistoricDateTime):
+            return value
```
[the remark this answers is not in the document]
iacobfred: Good point, I'll fix the logic.

alashow (commented):
LGTM

PR #577, opened 2021-08-07
MH-37: make topic tags clickable

iacobfred:
Resolves https://modularhistory.atlassian.net/browse/MH-37


frontend/components/propositions/OccurrenceDetail.tsx
```diff
@@ -39,8 +40,13 @@ const OccurrenceDetail: FC<OccurrenceDetailProps> = ({ occurrence }: OccurrenceD
...
       {occurrence.postscript && <p dangerouslySetInnerHTML={{ __html: occurrence.postscript }} />}
-      {occurrence.tagsHtml && (
-        <ul className="tags" dangerouslySetInnerHTML={{ __html: occurrence.tagsHtml }} />
+
+      {!!occurrence.cachedTags?.length && (
+        <ul className="tags">
+          {occurrence.cachedTags.map((topic) => (
+            <Tag key={topic.pk} topic={topic} />
+          ))}
+        </ul>
```
mattfredericksen, line 49: As you mentioned, we should probably pull the move the [`li` in `Tag`](https://github.com/ModularHistory/modularhistory/pull/577/files#diff-37d2c9b2992d497dcd7baef41fb6d104295f3bd41075d8f745037bc7d8e045adR21) out here so the usage of `li` is coupled with `ul`.

frontend/components/topics/Tag.tsx
```diff
@@ -0,0 +1,31 @@
...
+    backgroundColor: "rgba(255, 255, 153, 0.8)",
+    fontSize: "0.7rem",
+  },
+});
+
+interface TagProps {
+  topic: Topic;
+}
+
+const Tag: FC<TagProps> = ({ topic }: TagProps) => {
```
mattfredericksen, line 18: Do we prefer the name `Tag`? Or could we rename this component to `Chip` or `ChipLink` (or something else that might better convey what the component is).

iacobfred: I think I'd rather not name it `Chip` since I intended for the component to be used specifically for topic tags (not as a themed version of MUI's chip or something like that)—but, I don't have a strong opinion.

PR #632, opened 2021-08-31
Update MUI beta version

mattfredericksen:
Modified _document and _app to work with emotion. Other files were just ESLint fixes.


frontend/pages/_app.tsx
```diff
@@ -92,7 +95,7 @@ const App: NextPage<AppProps> = ({ Component, pageProps, err }: ExtendedAppProps
           </PageTransitionContextProvider>
         </SessionKiller>
       </Provider>
-    </>
+    </CacheProvider>
```
iacobfred, line 98: Do we still need `StyledEngineProvider` and `StylesProvider` above?

mattfredericksen: After reading the documentation for both, it looks like we should keep `StyledEngineProvider` while we're using JSS (which we should move away from going forward), but I don't think we need `StylesProvider`.

Just made some changes and tested on a local production build; SSR seems to work fine with JS disabled.

iacobfred (approved):
LGTM

PR #633, opened 2021-08-31
MH-224

mattfredericksen:
https://modularhistory.atlassian.net/browse/MH-224

When occurrences became a subtype of propositions, the index name logic broke here. ES was looking for a "propositions" index.


alashow (commented):
This is gonna work fine for now.

@iacobfred https://github.com/ModularHistory/modularhistory/commit/44d7f0024f44e8384816ef48d5362bce6b5af197 apparently changed es index name generation from static ContentType to static strings. I hope that was intentional?

iacobfred:
> @iacobfred [44d7f00](https://github.com/ModularHistory/modularhistory/commit/44d7f0024f44e8384816ef48d5362bce6b5af197) apparently changed es index name generation from static ContentType to static strings. I hope that was intentional?

Hey, yes, it was intentional, but I'm having trouble remember exactly why I did it. : - ) Maybe it was related to occurrences becoming a type of proposition. Are you OK with using hardcoded strings for the time being? (I'm thinking hardcoded strings are fine as we have a limited set of content types.)

alashow (approved):
@iacobfred Yeah I'm fine with hard-coded index names. Although we might want to add some kind of validation when parsing content_types in search to only allow known es indexes.

mattfredericksen:
> @iacobfred Yeah I'm fine with hard-coded index names. Although we might want to add some kind of validation when parsing content_types in search to only allow known es indexes.

That's a good idea, I'll add that soon.

PR #652, opened 2021-09-03
MH-235: zero-downtime deploy

iacobfred:


mattfredericksen (approved):
LGTM. I didn't look super closely at the bash logic though, but it seems like this has been tested well enough.

PR #654, opened 2021-09-03
Filter to valid content types during ES filter processing

mattfredericksen:
follow up for https://github.com/ModularHistory/modularhistory/pull/633#pullrequestreview-743279426


apps/search/api/filters/elastic_filters.py
```diff
@@ -36,7 +36,9 @@ def get_filter_params(request):
         indexes = '*'
         content_types = request.query_params.getlist('content_types') or None
         if content_types:
-            indexes = ','.join(content_types)
+            # #ContentTypesHardCoded
```
mattfredericksen, line 39: I don't know if this is a pattern we want to start using, but using "linked comments" like this was common at Asana to indicate when there is strong coupling to another part of the codebase.

iacobfred: No strong opinion

iacobfred (approved):

alashow (approved):

PR #657, opened 2021-09-05
MH-236: DRF serializers and ModelViewSets API's for moderated models

alashow:
Created DRF's [`ModelSerializer`](https://www.django-rest-framework.org/api-guide/serializers/#modelserializer) versions of moderated models serializers so it can be used with DRF [ModelViewSet](https://www.django-rest-framework.org/api-guide/viewsets/#modelviewset) API's. 

Serpy serializers were not suitable to use with ModelViewSet API's because it didn't support validation and it had some other integration errors. 

New `ModelSerializer`s were moved to `apps/{app}/api/serializers.py` module and serpy serializers will be located in `apps/{app}/serializers.py` module.

With the usage of ModelViewSet, now most moderated API's support CRUD operations through API (create/update operations only for authenticated users).
Create/update operations and how it works with Moderation models  have not been tested yet (although it's expected to work because ModelViewSet uses django model's `save()` method)


apps/entities/api/serializers.py
```diff
@@ -1,27 +1,21 @@
-"""Serializers for the entities app."""
+from rest_framework import serializers
 
-from typing import TYPE_CHECKING
+from apps.entities.models.entity import Entity
+from core.models.module import TypedModuleSerializerDrf
```
iacobfred, line 4: nit: Can we do `DrfBlahBlahSerializer` or `BlahBlahDrfSerializer` instead of `BlabBlahSerializerDrf`?

apps/entities/api/serializers.py
```diff
@@ -32,25 +26,18 @@ def get_serialized_end_date(self, instance: 'Entity'):
...
-        many=True, attr='categorizations.all', call=True
+        many=True, required=False, source='categorizations.all'
     )
 
-    def get_model(self, instance) -> str:  # noqa
-        """Return the model name of the instance."""
-        return f'entities.{instance.__class__.__name__.lower()}'
+    birth_date = serializers.DateField(write_only=True, required=False)
+    death_date = serializers.DateField(write_only=True, required=False)
+    birthDate = serializers.SerializerMethodField('get_serialized_birth_date', read_only=True)
```
iacobfred, line 39: Do we need `birth_date` and `birthDate` (same with `death_date` and `deathDate`)?

In our `REST_FRAMEWORK` settings, we specify automatically converting between snake case and camel case, so ideally we should use snake case in our Python code and camel case in our JavaScript code.
```
'DEFAULT_RENDERER_CLASSES': (
        # https://github.com/vbabiy/djangorestframework-camel-case
        'djangorestframework_camel_case.render.CamelCaseJSONRenderer',
        'djangorestframework_camel_case.render.CamelCaseBrowsableAPIRenderer',
        'rest_framework.renderers.JSONRenderer',
    ),
    'DEFAULT_PARSER_CLASSES': (
        # https://github.com/vbabiy/djangorestframework-camel-case
        'djangorestframework_camel_case.parser.CamelCaseFormParser',
        'djangorestframework_camel_case.parser.CamelCaseMultiPartParser',
        'djangorestframework_camel_case.parser.CamelCaseJSONParser',
    ),
```

apps/entities/api/serializers.py
```diff
@@ -32,25 +26,18 @@ def get_serialized_end_date(self, instance: 'Entity'):
         return instance.end_date.serialize() if instance.end_date else None
 
 
-class EntitySerializer(SearchableModelSerializer):
+class EntityModelSerializer(TypedModuleSerializerDrf):
```
iacobfred, line 29: Per https://github.com/ModularHistory/modularhistory/pull/657/files#r702437998 , maybe this one could be `EntityDrfSerializer`.

apps/moderation/serializers.py
```diff
@@ -3,17 +3,18 @@
...
-
-    class ModeratedModelSerializer(ModelSerializer):
+    class ModeratedModelSerializer(model_serializer):
         fields = serializers.ReadOnlyField(source='get_moderated_fields')
 
-        class Meta:
-            model = model_cls
-            include = model_cls.get_moderated_fields()
+        class Meta(model_serializer.Meta):
+            fields = model_serializer.Meta.fields + ['fields']
```
iacobfred, line 18: @alashow, what do you think about this?

```
        class Meta(model_serializer.Meta):
            fields = model_serializer.Meta.model.get_moderated_fields() + ['fields']
```
I'm considering the possibility of `model_serializer.Meta.fields` being different than `model_serializer.Meta.model.get_moderated_fields()`.

apps/places/models/base.py
```diff
@@ -53,7 +53,7 @@ def __str__(self) -> str:
         """Return the location's string representation."""
         return self.string
 
-    def save(self):
+    def save(self, *args, **kwargs):
```
iacobfred, line 56: Nice

apps/entities/api/serializers.py
alashow: Yes we need both, because input params and serialized version's aren't the same format.

Snake case version is the input and camel case version is serialized output, and they both will transformed to camel case version in the output. 
This way API client's only need to know about camel cased versions: `birthDate`, `deathDate` 
Isn't this the desired behavior? I we rename to `serialized_x_date`, the API clients will have to use `xDate` for inputs and `serializedXDate` when parsing outputs.

apps/entities/api/serializers.py
alashow: I added the word `Model` to reflect that they're backed by DRF's `ModelSerializer`, should I still rename to DRF?

apps/moderation/serializers.py
alashow: I'm not sure how would that work?
If we do that, `get_moderated_model_serializer` won't have any original fields. It will have "moderated fields" in `fields` field + an array of "moderated fields" ( which is a dict, moderatad field names as key and field info as value, and not compatible with `fields` array)

apps/entities/api/serializers.py
iacobfred: I think what's confusing to me is that since `birth_date` and `death_date` are converted to `birthDate` and `deathDate` anyway before Django sends a response to the frontend, I'm not sure which value is received by the frontend. Will the `birthDate` value received by the frontend be the write-only `DateField` or the read-only `SerializerMethodField`?

apps/moderation/serializers.py
iacobfred: Oh you're totally right. I was thinking that `get_moderated_fields` would return a list of strings (field names).

apps/entities/api/serializers.py
iacobfred: The current naming makes sense in context of DRF's `ModelSerializer`, but I think I would prefer the `DrfSerializer` suffix for extra clarity and consistency, considering that `ModelSerializer` doesn't necessary evoke DRF (at least in my mind) and we include `Drf` in the names of our parent classes for these serializers. Would you be OK with that?

apps/moderation/serializers.py
iacobfred: Since this serializer is for moderation, should we account for `model_serializer.Meta.model.Moderation.excluded_fields`? Maybe something like ...
```
fields = [
    field for field in model_serializer.Meta.fields 
    if field not in model_serializer.Meta.model.Moderation.excluded_fields
] + ['fields']
```

apps/entities/api/serializers.py
alashow: Since `birth_date/death_date` have `write_only=True`, they won't be rendered so frontend only receives serialized versions (`birthDate/deathDate` transformed to `birth_date/death_date` by DRF).

apps/moderation/serializers.py
alashow: I don't know why we would want to exclude fields for moderation? Do you have an example? 
`get_moderated_model_serializer` should just help extend normal serializers to add moderated fields info. 

Also, I haven't decided to use the function for moderation yet because we would want something that works with `ModelViewSet`s.

apps/entities/api/serializers.py
iacobfred: Awesome, OK, that makes sense. Could you please add a comment to the code about this?

apps/moderation/serializers.py
iacobfred: Actually, if we use `fields` for building the moderation form, then it's probably OK to not worry about limiting the fields in the serializer.  So, never mind.

But to answer your question, the idea of excluding fields for moderation came from the django-moderation package. And here are the default excluded fields for `ModeratedModel`:
```
class Moderation:
        excluded_fields = [
            'changes',
            'verified',
            'deleted',
            'cache',
            'pretty_cache',
            'date_string',
        ]
```

iacobfred (approved):
LGTM!

apps/search/api/serializers.py
```diff
@@ -27,3 +22,24 @@ def get_meta(self, model) -> Optional[dict]:
...
+
+    def get_meta(self, model) -> Optional[dict]:
+        if not hasattr(model, 'meta'):
+            return None
+        meta = model.meta.to_dict()
+        for key in ELASTICSEARCH_META_FIELDS_TO_CLEAN:
+            # TODO: find out why in some cases it's already cleaned
+            if key in meta:
+                del meta[key]
+        return model.meta.to_dict()
```
mattfredericksen, line 40: Not sure if I'm missing something here, but maybe the logic is broken? Assuming `to_dict()` is just returning a dictionary representation of `model.meta`, is it possible that deleting keys on that dictionary is not mutating `model.meta`? Then we just return `to_dict()` again, so maybe nothing has been modified. Do we want to `meta` instead of calling `to_dict` again?

mattfredericksen, line 40: Not sure if I'm missing something here, but maybe the logic is broken? Assuming `to_dict()` is just returning a dictionary representation of `model.meta`, is it possible that deleting keys on that dictionary is not mutating `model.meta`? 
Then we just return `to_dict()` again, so maybe nothing has been modified. Do we want to return `meta` instead of calling `to_dict` again?

mattfredericksen (changes requested):
LGTM! Just had one concern with `get_meta`.

apps/search/api/serializers.py
alashow: Good catch! First I couldn't understand why this is actually working (i.e cleans given keys). 
After reading some `elasticsearch-dsl` code, it looks like [`to_dict()`](https://github.com/elastic/elasticsearch-dsl-py/blob/7874040b4f1346552a77b6f2a16321f2e1a6c722/elasticsearch_dsl/utils.py#L196-L197) actually returns it's internal `dict` instead returning a new dict. 

So when we're mutating `meta`, we're mutating `model.meta.to_dict()` to because they both reference to same dictionary (tbh, I'm assuming this). This also explains `TODO: find out why in some cases it's already cleaned`

I'm wondering whether we need do to clone the returned dict and return `meta` to make it cleaner.

iacobfred (approved):

mattfredericksen (approved):
LGTM!

PR #659, opened 2021-09-05
MH 68: Capitalize the first word of occurrences

senamn:


frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -8,6 +8,46 @@ interface ModuleUnionCardProps {
   selected?: boolean;
 }
 
+function titleCase(word: string) {
```
mattfredericksen, line 11: Hey Sena, are you familiar with using `npm`? This feature seems to me like a good candidate for installing a package instead of writing/maintaining our own code.

frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -22,7 +62,7 @@ const ModuleUnionCard: FC<ModuleUnionCardProps> = ({
       content = <HTMLEllipsis unsafeHTML={module.captionHtml} maxLine="3" basedOn="words" />;
       break;
     case "propositions.occurrence":
-      content = <div dangerouslySetInnerHTML={{ __html: module.title }} />;
+      content = <div dangerouslySetInnerHTML={{ __html: titleCase(module.title) }} />;
```
mattfredericksen, line 65: I think it's possible we don't need `dangerouslySetInnterHTML` for titles anymore, since they might all be raw text (no HTML tags). But if that's wrong, have we confirmed the case-changing logic does not affect HTML tags?

frontend/components/cards/ModuleUnionCard.tsx
senamn: Hey Matthew, I'm not too familiar with the application of a package to our project but I was able to find a few packages that seemed promising such as:

https://www.npmjs.com/package/title-case

mattfredericksen: Yeah, I noticed that one first, but I also noticed [`change-case`](https://www.npmjs.com/package/change-case), which:
1. has roughly the same popularity (weekly downloads) so it probably works well,
2. has more available formats for case translations ([including title case](https://github.com/blakeembrey/change-case#titlecase)), and
3. is surprisingly a smaller package (9.33kb vs 27.7kb).

We should be minimizing package size when that's an option, so I'd prefer that we go with `change-case`, assuming the title case function works well.

To add the package, you'll need to navigate to the `frontend/` directory and run `npm i <package-name>`. Then you can import it where it's needed, following the documentation on the npm/GitHub pages.

frontend/components/cards/ModuleUnionCard.tsx
senamn: Oh, I see. Do you suggest that I remove `dangerouslySetInnerHTML` and `__html` and test if it runs fine? And affecting HTML tags in which way do you mean?

frontend/components/cards/ModuleUnionCard.tsx
senamn: Got ya, thanks for letting me know! I'll test this tonight and let you know how it goes.

frontend/components/cards/ModuleUnionCard.tsx
iacobfred: Good point @mattfredericksen. @senamn, you could test it like this:
```
content = <div>{titleCase(module.title)}</div>;
```

mattfredericksen: To make sure all our data is safe, I would take a look at where that field is defined in our Django models, and maybe run a query like `Propositions.objects.all().values('title')`, and see if the output contains any html.

frontend/components/cards/ModuleUnionCard.tsx
senamn: @mattfredericksen I was able to download `change-case` package and apply it but it seems that `titleCase` is not included and need to be installed separately. I'll try to look for a title case package that is small and has many downloads. In the meantime, to remove the package I downloaded, is there anything else I need to worry about other than removing related lines from package-lock.json and package.json?

mattfredericksen: Ah, I didn't notice that message in the docs. I think the first package you mentioned is fine then, I wouldn't worry about the size since it seems like the best option.

mattfredericksen: >In the meantime, to remove the package I downloaded, is there anything else I need to worry about other than removing related lines from package-lock.json and package.json?

I removing the lines from `package.json` works fine, but you can also use `npm remove <package>`.

senamn: > Ah, I didn't notice that message in the docs. I think the first package you mentioned is fine then, I wouldn't worry about the size since it seems like the best option.

Sounds good!

senamn: > > In the meantime, to remove the package I downloaded, is there anything else I need to worry about other than removing related lines from package-lock.json and package.json?
> 
> Removing the lines from `package.json` works fine, but you can also use `npm remove <package>`.

Oh, got it. Thanks for letting me know!

frontend/components/cards/ModuleUnionCard.tsx
senamn: I found one title that includes html but couldn't find an occurrence under this title:
Screen Shot 2021-09-07 at 12 30 47 AM

Also, while searching for that <em>Homo sapiens</em>, I noticed this is what is shown in production:
Screen Shot 2021-09-07 at 12 33 01 AM

Other than that, there wasn't any html included in the title.

iacobfred: @senamn thanks for checking! If there's just one instance, I can just fix the data in production so that the title doesn't contain HTML.

senamn: > @senamn thanks for checking! If there's just one instance, I can just fix the data in production so that the title doesn't contain HTML.

No worries and got it! So do you want me to wait until you fix it or just go ahead and push the updated code? And do you want me to change module.title line to be `content = <div>{titleCase(module.title)}</div>;`?

iacobfred: You can go ahead and push the updated code, and yes, that should work!

frontend/components/propositions/OccurrenceDetail.tsx
```diff
@@ -36,7 +37,10 @@ const OccurrenceDetail: FC<OccurrenceDetailProps> = ({ occurrence }: OccurrenceD
           )
       )}
 
-      <h2 className="text-center my-3" dangerouslySetInnerHTML={{ __html: occurrence.title }} />
+      <h2
+        className="text-center my-3"
+        dangerouslySetInnerHTML={{ __html: titleCase(occurrence.title) }}
+      />
```
mattfredericksen, line 43: [Same as with the `div`](https://github.com/ModularHistory/modularhistory/pull/659/files#r702441243), we should be able to replace this with 
```jsx
<h2 className="...">
  titleCase(...)
</h2>
```

mattfredericksen (changes requested):

frontend/components/propositions/OccurrenceDetail.tsx
```diff
@@ -36,7 +37,10 @@ const OccurrenceDetail: FC<OccurrenceDetailProps> = ({ occurrence }: OccurrenceD
...
       {occurrence.summary != occurrence.title && <p className="lead">{occurrence.summary}</p>}
```
iacobfred, line 44: @senamn, can you change this line so that the `p` element uses `dangerouslySetInnerHTML` for `occurrence.summary`? 

That will fix the <em> tags being displayed around "Homo sapiens."

frontend/components/propositions/OccurrenceDetail.tsx
senamn: > [Same as with the `div`](https://github.com/ModularHistory/modularhistory/pull/659/files#r702441243), we should be able to replace this with
> 
> ```js
> <h2 className="...">
>   titleCase(...)
> </h2>
> ```

Got ya. Fixed it and will be pushing the updated code!

frontend/components/propositions/OccurrenceDetail.tsx
senamn: > @senamn, can you change this line so that the `p` element uses `dangerouslySetInnerHTML` for `occurrence.summary`?
> 
> That will fix the `<em>` tags being displayed around "Homo sapiens."

Yes, fixed it accordingly and will be pushing the updated code!

iacobfred:
LGTM!

iacobfred (approved):

mattfredericksen (approved):
Seeing proper capitalization makes me unreasonably happy :)
LGTM!

PR #664, opened 2021-09-06
MH-233

mattfredericksen:
For those who joined the [pair-programming](https://unt.zoom.us/rec/share/k1I96448SZxQYDeVSI8FRFjPJjBJrOEXC1dqb7eAgOkUC7wra9lnDfRWux-l5Jwu.2FauaVupQwzwueqp), after the recording ended I:
- moved the component from `Layout` to `App`, since `Layout` re-renders every time the page changes (since the `title` prop changes), and re-renders reset state (i.e. transition progress), which would prevent the gradual fade out on load completion.
- replaced the funky `setTimeout` stuff with [`setInterval`](https://developer.mozilla.org/en-US/docs/Web/API/setInterval), which I had forgotten about.
- added `timerIDRef`, allowing us to prevent consecutive page transitions from overwriting each other with state updates.
- figured out the CSS stuff, the bar was disappearing since we didn't have `width={"100%"}`, I think.
- moved the component definition to a distinct file, allowing us to dynamically import it after initial page load, reducing the first-load page size (this is probably something we should always do when a component does not affect SSR).

https://user-images.githubusercontent.com/24942155/132164053-94e0e5a3-540c-4ea1-9ab0-948f6ad11e93.mp4


nickghuman (approved):

frontend/components/PageTransitionProgressBar.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    return () => {
+      events.off("routeChangeStart", handleRouteChangeStart);
+      events.off("routeChangeComplete", handleRouteChangeComplete);
+    };
+  }, [events]);
+
+  return (
+    <Box position={"fixed"} width={"100%"} top={-1} left={0} zIndex={10}>
+      <Fade in={![0, 100].includes(loadingProgress)} timeout={{ exit: 1e3 }}>
```
alashow, line 50: Just interested, what does `<Fade ... timeout={{ exit: 1e3 }}>` mean? 1000ms fade out duration?

alashow (approved):
LGTM

frontend/components/PageTransitionProgressBar.tsx
mattfredericksen: Exactly. I tried a few different fade out durations and that one seemed best. The fade in default duration seemed fine.

PR #676, opened 2021-09-07
Reduce ES memory limits in dev

mattfredericksen:
We experienced some crashes before when lowering this value, but everything seems to work fine now. It might be good to test this change locally before approving, but it's also easy to fix if it causes problems.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -101,6 +101,7 @@ services:
       - "9200:9200"
     environment:
       - ELASTIC_PASSWORD=test
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
 
   kibana:
     ports:


docker-compose.override.yml
```diff
@@ -101,6 +101,7 @@ services:
       - "9200:9200"
     environment:
       - ELASTIC_PASSWORD=test
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
```
mattfredericksen, line 104: https://github.com/ModularHistory/modularhistory/blob/6ff333233af734f9212f6c1f6175801d78750f45/docker-compose.yml#L162

mattfredericksen: @alashow it limits the heap size, which does not include statically allocated memory. For me, it reduced memory consumption from ~2G to ~1G.

PR #689, opened 2021-09-09
MH-240: Enable SERP module anchors

mattfredericksen:
Unfortunately, we can't just attach ids to each module card (the way url anchors/hashes usually work), since then updating the hash of the page causes the left pane to jump to the selected module. A manual implementation was needed.

https://user-images.githubusercontent.com/24942155/132877905-76146d92-762b-491b-8fbc-99eb987b6f72.mp4

Note: after recording I updated the behavior so if the url anchor links to the first module then it does not scroll down.


aaw0151 (approved):
Looks good. Great job!

nickghuman (approved):

PR #690, opened 2021-09-09
MH-66: Implemented content snippets for all result cards. Changed lay…

aaw0151:
…out of title cards to be more uniform.


frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -22,22 +23,52 @@ const ModuleUnionCard: FC<ModuleUnionCardProps> = ({
       content = <HTMLEllipsis unsafeHTML={module.captionHtml} maxLine="3" basedOn="words" />;
       break;
     case "propositions.occurrence":
-      content = <div dangerouslySetInnerHTML={{ __html: module.title }} />;
+      content = module.meta &&
+        module.meta.highlight &&
+        module.meta.highlight.text &&
+        module.meta.highlight.text.length > 0 && ( // TODO: Replace Regex with more formal fix
```
mattfredericksen, line 29: We should be able to simplify this to `module.meta?.highlight?.text && (...)`.

frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -22,22 +23,52 @@ const ModuleUnionCard: FC<ModuleUnionCardProps> = ({
...
-        </blockquote>
+        //<blockquote className="blockquote">
+        <div>
+          {(module.meta &&
+            module.meta.highlight &&
+            module.meta.highlight.text &&
+            module.meta.highlight.text.length > 0 && ( // TODO: Replace Regex with more formal fix
+              <Box sx={{ "& *": { fontSize: "0.8rem", backgroundColor: "rgba(0, 0, 0, 0)" } }}>
+                <HTMLEllipsis
+                  unsafeHTML={module.meta.highlight.text[0].replace(/(<([^>]+)>)/gi, "")}
```
iacobfred, line 63: It looks like the regex pattern here removes the HTML tags but leaves their contents, which _mostly_ works but unfortunately has weird results in some cases. For example, if there's a header tag captured in the highlight, simply stripping out all the HTML tags results in something like this (notice the "Adam See" at the beginning):

adam see

It also would be nice to do something with the <mark> tags (which contains the text that matched the search). Since highlighting doesn't look good on the card, maybe we could apply a **bold** style instead?

^ So, considering both of these things, I think we should:
- Change the regex pattern to remove header tags _and their contents_ (i.e., in the example above, remove the initial "Adam" altogether) but leave other HTML tags present
- Add an additional style rule for `mark` tags to make them bold—like this:
```
sx={{ "& *": { fontSize: "0.8rem", backgroundColor: "rgba(0, 0, 0, 0)" }, "& mark": { fontWeight: "bold" } }}
```

aaw0151: Oh my mistake, I didn't know that's what you meant. I'll update the regexs!

frontend/types/modules.d.ts
```diff
@@ -40,14 +42,25 @@ export interface ModuleWithImages {
   cachedImages: Image[];
 }
 
+export interface Text {
+  text?: string[]; //for quotes.quote
```
iacobfred, line 46: nit: Can you add a space after the `//` for your comments?

frontend/types/modules.d.ts
```diff
@@ -40,14 +42,25 @@ export interface ModuleWithImages {
...
+  highlight: Text; //array of Text object
+  sort: number[];
+}
+
 export interface Quote extends SearchableModule, ModuleWithImages {
   model: "quotes.quote";
   title: string;
   attributeeHtml: string;
-  attributeeString: string;
+  attributeeString: string; //name of person who is quoted
```
mattfredericksen, line 59: IMO we should only put comments describing data model fields in Django's model definitions, but I'll leave that decision to others.

mattfredericksen (changes requested):

frontend/components/cards/ModuleUnionCard.tsx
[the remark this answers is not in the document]
mattfredericksen: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#description

frontend/components/cards/ModuleCard.tsx
```diff
@@ -169,10 +169,26 @@ const ModuleCard: FC<ModuleCardProps> = ({
...
+          <small dangerouslySetInnerHTML={{ __html: module.dateString }} />
+        </CardHeader>
+      ) : module.name ? ( // if just name
+        <CardHeader className={"text-center"}>
+          <small dangerouslySetInnerHTML={{ __html: module.name }} />
+        </CardHeader>
+      ) : module.title ? ( // if just title
+        <CardHeader className={"text-center"}>
+          <small dangerouslySetInnerHTML={{ __html: module.title }} />
+        </CardHeader>
```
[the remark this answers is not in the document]
aaw0151: Unfortunatly, module.name and module.title are not the same. Depending on the type of card, it may have title or name, or neither. But your implementation does look better and cleaner so I will be adding that right now.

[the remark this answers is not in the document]
mattfredericksen: Do you have an example of this? I thought that all module with `name` also have `title` (e.g. [entities `title` is set to its `name`](https://github.com/ModularHistory/modularhistory/blob/601bebd7088363bc315482ff1d3af35e59a76cd8/apps/entities/models/entity.py#L183)).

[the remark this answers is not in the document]
aaw0151: image
This is using just name.
image
This is using just title.
image
This is using both.

Some have just title and some have just name. I may have been mistaken about some cards having neither.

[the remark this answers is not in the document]
iacobfred: Oh interesting. Entity's `title` is supposed to be set from its `name`, but the `title` field was added after `name` and does not get set until the object is saved. All our current entities have not been re-saved since the `title` field was added, so they don't have `title` values yet. Hopefully that makes sense!

I'll run a script in production shortly to set the `title` values from `name`.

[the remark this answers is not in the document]
aaw0151: I'll check it out now, thanks!

frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -22,22 +42,25 @@ const ModuleUnionCard: FC<ModuleUnionCardProps> = ({
       content = <HTMLEllipsis unsafeHTML={module.captionHtml} maxLine="3" basedOn="words" />;
       break;
     case "propositions.occurrence":
-      content = <div dangerouslySetInnerHTML={{ __html: module.title }} />;
+      content = module?.meta?.highlight?.text && ( // TODO: Replace Regex with more formal fix
```
iacobfred, line 45: Looks like the `module?.meta?.highlight?.text` check appears multiple times in this file. How about putting it in a variable before the `switch` statement?
```
const snippet = module?.meta?.highlight?.text?.[0]
```
Then, the content assignments could be made more concise like this:
```
      content = snippet && (
        <HighlightEllipsis unsafeHTML={snippet} />
      );
```

aaw0151: Good idea! Its been pushed!

iacobfred (approved):
LGTM!

frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -16,28 +36,33 @@ const ModuleUnionCard: FC<ModuleUnionCardProps> = ({
   // ModuleUnionCard is a generic component for rendering cards of any
   // model type, removing the need to import every card component.
   let content;
+  const highlightSnippet =
+    module?.meta?.highlight?.text?.[0] ?? module?.meta?.highlight?.description?.[0]; // highlight text or highlight description
```
mattfredericksen, line 40: I'm not sure exactly how the highlight API works, but we use `text` here because that field exists on quotes, and `description` because that field exists on entities, right? If so, do we need `elaboration` for propositions/occurrences? And I'm not sure what the fields would be for images and sources, if we want those highlighted as well.

mattfredericksen (changes requested):

frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -48,7 +73,9 @@ const ModuleUnionCard: FC<ModuleUnionCardProps> = ({
     case "entities.organization":
     case "entities.entity":
     case "entities.group":
-      content = <div dangerouslySetInnerHTML={{ __html: module.name }} />;
+      content = highlightSnippet && ( // TODO: Replace Regex with more formal fix
+        <HighlightEllipsis unsafeHTML={highlightSnippet} />
+      );
```
iacobfred, line 78: How about adding a fallback value in case there's no `highlightSnippet`?

Something like...
```
      content = highlightSnippet && (
        <HighlightEllipsis unsafeHTML={highlightSnippet} />
      ) || (
        <HTMLEllipsis unsafeHTML={entity.description} maxLine="4" basedOn="words" />
      )
```

frontend/components/cards/ModuleUnionCard.tsx
iacobfred: In `apps/search/api/filters/elastic_filters.py`, I found this code:
```
        # TODO: refactor & improve this. currently only applying highlights to quote#text and occurrence#description
        qs = qs.highlight(
            'text',
            number_of_fragments=1,
            type='plain',
            pre_tags=['<mark>'],
            post_tags=['</mark>'],
        )
        qs = qs.highlight(
            'description',
            number_of_fragments=1,
            type='plain',
            pre_tags=['<mark>'],
            post_tags=['</mark>'],
        )
```

For now, we could just just add `elaboration` to this like so...
```
        qs = qs.highlight(
            'text',
            number_of_fragments=1,
            type='plain',
            pre_tags=['<mark>'],
            post_tags=['</mark>'],
        ).highlight(
            'description',
            number_of_fragments=1,
            type='plain',
            pre_tags=['<mark>'],
            post_tags=['</mark>'],
        ).highlight(
            'elaboration',
            number_of_fragments=1,
            type='plain',
            pre_tags=['<mark>'],
            post_tags=['</mark>'],
        )
```

frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -8,6 +9,25 @@ interface ModuleUnionCardProps {
   selected?: boolean;
 }
 
+interface HighlightEllipsisProps {
+  unsafeHTML: string;
+}
+
+const HighlightEllipsis: FC<HighlightEllipsisProps> = ({ unsafeHTML }) => (
+  <Box
+    sx={{
```
iacobfred, line 18: I've noticed that every once in a while, we still get unexpected HTML elements with block display (e.g., `p` tags). Here's an example:
Screen Shot 2021-09-11 at 12 17 44

I think for the content snippets, it would be best if we displayed the content inline (i.e., with `display: inline`, but included ellipses so that the content isn't jammed together incomprehensibly. Something like this:
Screen Shot 2021-09-11 at 12 30 13

frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -8,6 +9,25 @@ interface ModuleUnionCardProps {
...
+      "& *": { fontSize: "0.8rem", backgroundColor: "rgba(0, 0, 0, 0)" },
+      "& mark": { fontWeight: "500", color: "white" },
+    }}
+  >
+    <HTMLEllipsis
+      unsafeHTML={unsafeHTML.replace(/(<(h[1-6])>.*?<\/h[1-6]>)/gi, "")}
```
iacobfred, line 24: Elasticsearch sometimes gives us hightlights content that contains empty `p` tags, i.e., `<p></p>`.
So, maybe we should add another regex replacement to get rid of those. (I looked briefly in the elasticsearch documentation and didn't see a way to avoid this from the backend.)

The additional rule for remove empty tags could be `.replace(/<(p|div)><\/(p|div)>/gi, "")` (May as well remove both empty `p` tags and empty `div` tags, although I haven't encountered empty divs yet.)

iacobfred (changes requested):
See https://github.com/ModularHistory/modularhistory/pull/690/files#r706638945

frontend/components/cards/ModuleUnionCard.tsx
aaw0151: Not sure if that fixed the problem as I was still seeing some leading ellipsis in content highlights, but I may fix that in the next sprint in a short task. Your fix has been implemented though.

frontend/components/cards/ModuleUnionCard.tsx
aaw0151: Added fallbacks to all content types.

frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -8,6 +9,27 @@ interface ModuleUnionCardProps {
...
 }
 
+interface HighlightEllipsisProps {
+  unsafeHTML: string;
+}
+
+const HighlightEllipsis: FC<HighlightEllipsisProps> = ({ unsafeHTML }) => (
+  <Box
+    sx={{
+      "& *": { fontSize: "0.8rem", backgroundColor: "rgba(0, 0, 0, 0)" },
```
iacobfred, line 19: @aaw0151, can you also update this line to render all elements inline?  Like so:

`"& *": { fontSize: "0.8rem", backgroundColor: "rgba(0, 0, 0, 0)", display: "inline" },`

iacobfred (approved):
LGTM!

frontend/types/modules.d.ts
```diff
@@ -33,13 +35,25 @@ export interface Image extends SearchableModule {
...
+  text?: string[]; // for quotes.quote
+  description?: string[]; // for sources.source, entities.person
+  elaboration?: string[]; // for occurances
+}
+export interface Meta {
+  score: number;
+  highlight: Text;
+  sort: number[];
+}
+
```
mattfredericksen, line 56: nit: Just for consistency we should eventually have a newline between these interface definitions, but that doesn't have to be part of this PR.

mattfredericksen (approved):
LGTM!

PR #733, opened 2021-09-17
poetry PATH fix

mattfredericksen:
At least in my environment, and maybe in others, poetry commands in our git hooks were not running since `poetry` was not in `PATH`. I think this is because my IDE does not currently run the hooks with the virtual environment activated.

### DIFF
--- a/config/hooks/post-merge
+++ b/config/hooks/post-merge
@@ -1,4 +1,5 @@
 #!/bin/bash
+source "$HOME/.profile"
 
 # For debugging:
 echo "Successfully pulled to $(pwd)."
--- a/config/hooks/pre-commit
+++ b/config/hooks/pre-commit
@@ -19,22 +19,23 @@ echo "$staged_files" | grep -E --quiet ".+.py" && {
     # Autoformat and lint Python code.
     poetry --quiet && [[ -d ".venv" ]] && {
         comma_delimited_filepaths=$(echo "$staged_python_files" | tr '\n' ',')
+        space_delimited_filepaths=$(echo "$staged_python_files" | tr '\n' ' ')
         echo "Autoformatting Python code..."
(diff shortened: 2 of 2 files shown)


config/hooks/post-merge
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+source "$HOME/.profile"
```
mattfredericksen, line 2: We could update `PATH` just for poetry here, but I don't think sourcing `.profile` will cause any problems, and it seems better for avoiding similar problems in the future.
```bash
export PATH="$HOME/.poetry/bin:$PATH"
```

config/hooks/pre-commit
```diff
@@ -16,24 +17,25 @@ echo "Linting Python code..."
...
+        poetry run mypy "$space_delimited_filepaths" || {
+            echo "
+                ${BOLD}Examine the error above. Should it be fixed?${NORMAL}
+                Pytype may give more helpful output:
+                    poetry run pytype --output .cache/pytype $file
+            "
+        }
+
+
+        poetry run flake8 "$space_delimited_filepaths" || linting_failed=true
```
mattfredericksen, line 37: I'm not sure if there was a specific reason to run these for each file, but since the checks run recursively through imports, it was taking forever to run on individual files. This should deduplicate that work.

config/hooks/pre-commit
```diff
@@ -16,24 +17,25 @@ echo "Linting Python code..."
...
+
```
mattfredericksen, line 38: I'm not sure if there was a specific reason to re-run these for each file, but since the checks run recursively through imports, it was taking forever to run on individual files. This should deduplicate that work.

mattfredericksen:
This can't break anything in production, so I'm going to go ahead and merge, and if it breaks something in dev we can revert.

PR #742, opened 2021-09-20
Mh 246

aaw0151:


frontend/pages/topics/index.tsx
```diff
@@ -24,16 +53,11 @@ const Topics: FC<TopicsProps> = ({ topicsData }: TopicsProps) => {
...
-                  display: "inline-block",
-                }}
-              >
-                {topic.name}
-              </Button>
+              <StyledTopicCard className={`m-2 ${className || ""}`} data-type={topic.model}>
+                <CardContent>
+                  <h3>{topic.name}</h3>
+                </CardContent>
+              </StyledTopicCard>
```
mattfredericksen, line 60: I'm not sure this is an improvement over the previous layout. The cards are pretty bulky now, and the previous buttons had some nice hover effects that are removed here. Maybe `<Card/>` is not the best type of component for this page?

Do you plan to refine the design here more? I don't think this should be merged as-is.

frontend/pages/topics/[slug].tsx
```diff
@@ -22,12 +68,19 @@ const TopicDetailPage: FC<TopicProps> = ({ topic }: TopicProps) => {
         {topic.description && <p dangerouslySetInnerHTML={{ __html: topic.description }} />}
         {topic.propositions &&
           topic.propositions.map((proposition) => (
-            <Link href={`/propositions/${proposition.slug}`} key={proposition.slug}>
-              <a>
```
Comment (mattfredericksen) [inline frontend/pages/topics/[slug].tsx:26]: We probably don't want to remove this anchor. [`Link`](https://nextjs.org/docs/api-reference/next/link)'s documentation mentions that SEO might be harmed by not rendering an anchor, I think because the `href` doesn't appear otherwise.

mattfredericksen (changes requested):
Not a big deal, but it's usually helpful to include before/after screenshots for UI changes. 

About the new design: we might want to get more input/ideas from the team, or we could get more ideas by browsing MUI's component library.

frontend/pages/topics/index.tsx
aaw0151: I see what you mean. I tried to make it easier to find a topic that you are looking for. The current Topics page is rather filled and hard to find what you want. Maybe a search implementation on the page would work well? What do you think @mattfredericksen ?

aaw0151:
Here's a comparison of the previous topic results page and the proposed topic results page.
Before:
image

After:
image

frontend/pages/topics/index.tsx
mattfredericksen: I agree, the current page also has problems...

What do you think about adding dividers between each letter? I think it would be _an_ improvement, but I'm not sure if it's what you're looking for.

I don't have any other ideas off the top of my head, but typically its possible to find UI design examples for similar pages/features on other sites, which takes less effort and is usually higher quality than designs we could create ourselves, IMO. I would just run ideas by the team and/or Jacob before implementing anything non-trivial.

aaw0151:
This is the proposed changes to the Topics page.
Before:
image

After:
image

frontend/pages/topics/[slug].tsx
```diff
@@ -22,11 +36,24 @@ const TopicDetailPage: FC<TopicProps> = ({ topic }: TopicProps) => {
         {topic.description && <p dangerouslySetInnerHTML={{ __html: topic.description }} />}
         {topic.propositions &&
           topic.propositions.map((proposition) => (
-            <Link href={`/propositions/${proposition.slug}`} key={proposition.slug}>
+            <Link href={`/propositions/${proposition.slug}`} key={proposition.slug} passHref>
               <a>
-                <div>
-                  <p dangerouslySetInnerHTML={{ __html: proposition.summary }} />
-                </div>
+                <StyledTopicResultCard>
```
Comment (mattfredericksen) [inline frontend/pages/topics/[slug].tsx:41]: Should we add the [`raised`](https://mui.com/api/card/#props) prop here? I think it looks better.

frontend/pages/topics/[slug].tsx
```diff
@@ -3,9 +3,23 @@ import ModuleContainer from "@/components/details/ModuleContainer";
...
 import { Topic } from "@/types/modules";
+import { Card, CardContent, CardHeader } from "@material-ui/core";
+import { styled } from "@material-ui/core/styles";
 import { GetStaticPaths, GetStaticProps } from "next";
 import Link from "next/link";
-import { FC } from "react";
+import React, { FC } from "react";
+import HTMLEllipsis from "react-lines-ellipsis/lib/html";
+
+const StyledTopicResultCard = styled(Card)({
```
Comment (mattfredericksen) [inline frontend/pages/topics/[slug].tsx:13]: Nit: can we rename `StyledTopicResultCard` to `PropositionCard` or `TopicPropositionCard`? IMO it's best to prefix with `Styled` when it's not convenient to think of a more descriptive name (e.g. [`<StyledPagination/>`](https://github.com/ModularHistory/modularhistory/blob/2337d77001d0af0d25fadce3e5a7e3c9b6d13cba/frontend/components/Pagination.tsx#L7)), but otherwise `Styled` doesn't need to be in the name (e.g. [`<TagChip/>`](https://github.com/ModularHistory/modularhistory/blob/2337d77001d0af0d25fadce3e5a7e3c9b6d13cba/frontend/components/topics/Tag.tsx#L7)).

frontend/pages/topics/index.tsx
```diff
@@ -1,40 +1,57 @@
...
+  module: ModuleUnion;
+  className?: string;
 }
 
 const Topics: FC<TopicsProps> = ({ topicsData }: TopicsProps) => {
   const topics = topicsData.topics || [];
+  const topicsByLetters: Record<string, typeof topics> = {};
+  topics.map((topic) => {
+    const firstLetter = topic.name[0].toUpperCase() || " ";
+    topicsByLetters[firstLetter] = topicsByLetters[firstLetter] || []; //create array if not created already
```
mattfredericksen, line 24: I think we can remove the `|| " "` here.

Also nit: can we include a space after the `//` in the comment?

frontend/pages/topics/index.tsx
```diff
@@ -1,40 +1,57 @@
...
-                  display: "inline-block",
-                }}
-              >
-                {topic.name}
-              </Button>
-            </Link>
+          {Object.keys(topicsByLetters).map((key) => (
+            <>
+              <Divider sx={{ my: "2" }}>{key}</Divider>
+              {topicsByLetters[key].map((topic: any) => (
```
mattfredericksen, line 36: We probably don't want to cast `topic` to `any` here.

mattfredericksen (changes requested):
LGTM! Just a couple small changes requested.

frontend/pages/topics/index.tsx
```diff
@@ -1,40 +1,57 @@
...
-                  margin: "0.6rem 1.2rem",
-                  display: "inline-block",
-                }}
-              >
-                {topic.name}
-              </Button>
-            </Link>
+          {Object.keys(topicsByLetters).map((key) => (
+            <>
+              <Divider sx={{ my: "2" }}>{key}</Divider>
```
mattfredericksen, line 35: Oh, it looks like `"2"` is different than `2` here (so I think we should use `2`). 
https://mui.com/system/basics/#shorthands

mattfredericksen (changes requested):

mattfredericksen (approved):

PR #746, opened 2021-09-22
Dependency updates

mattfredericksen:
No need to look over the files, almost all of the changes are just find & replace.


setup.sh
```diff
@@ -311,7 +311,7 @@ poetry_init='export PATH="$HOME/.poetry/bin:$PATH"'
 poetry --version &>/dev/null || {
   echo "Installing Poetry ..."
   # https://python-poetry.org/docs/#osx-linux-bashonwindows-install-instructions
-  curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3 -
+  curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python3 -
```
mattfredericksen, line 314: image

mattfredericksen:
Tests were passing right before the poetry script update, and the failure there is unrelated to that change.

PR #747, opened 2021-09-22
topics instant search

mattfredericksen:
Adds instant search support for topics in the search form.

https://user-images.githubusercontent.com/24942155/134750872-dc9ad81f-7563-4b3a-8bac-517a52856b40.mp4

### DIFF
--- a/frontend/components/search/InstantSearch.tsx
+++ b/frontend/components/search/InstantSearch.tsx
@@ -124,7 +124,6 @@ const InstantSearch: FC<InstantSearchProps> = ({
       renderInput={(params) => <TextField {...params} variant="outlined" label={label} />}
       sx={{
         "& .MuiChip-root": {
-          margin: 1,
           fontSize: "11px",
           height: "20px",
           "& .MuiChip-label": {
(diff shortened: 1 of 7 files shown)


frontend/components/search/InstantSearch.tsx
```diff
@@ -124,7 +124,6 @@ const InstantSearch: FC<InstantSearchProps> = ({
       renderInput={(params) => <TextField {...params} variant="outlined" label={label} />}
       sx={{
         "& .MuiChip-root": {
-          margin: 1,
```
mattfredericksen, line 127: This was adding a lot of space between chips.
>image

aaw0151 (approved):
Looks good to me! Great job! Will probably be using some of this if we want a similar dynamic search display for the topics page.

mattfredericksen:
> Looks good to me! Great job! Will probably be using some of this if we want a similar dynamic search display for the topics page.

Sounds interesting! I plan to abstract the logic in `<InstantSearch/>` at some point so we can easily reuse it in other areas of the app.

PR #756, opened 2021-09-25
MH-203: Added Alert MUI Component to Success/Error Information

sunzae99:


mattfredericksen:
It looks like some changes here were duplicated in #757.
`main` needs to be merged in, and this error needs to be fixed.
image

PR #760, opened 2021-09-25
MH-241: Updated fallback content snippets and removed uneeded HTML ta…

aaw0151:
…gs from snippets.

### DIFF
--- a/frontend/components/cards/ModuleUnionCard.tsx
+++ b/frontend/components/cards/ModuleUnionCard.tsx
@@ -8,9 +8,9 @@ interface HighlightEllipsisProps {
   unsafeHTML: string;
 }
 
-function applyHTMLFilter(str: string) {
+function applyHTMLFilter(unsafeHTML: string) {
   // TODO: Replace regex with more formal fix
-  return str.replace(
+  return unsafeHTML.replace(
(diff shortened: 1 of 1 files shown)


frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -51,41 +52,61 @@ const ModuleUnionCard: FC<ModuleUnionCardProps> = ({
...
       content = (
-        <blockquote className="blockquote">
-          <div>
-            {(highlightSnippet && <HighlightEllipsis unsafeHTML={highlightSnippet} />) ??
-              (module.bite && (
-                <HTMLEllipsis unsafeHTML={module.bite} maxLine="3" basedOn="words" />
-              ))}
-          </div>
-        </blockquote>
+        //<blockquote className="blockquote">
```
mattfredericksen, line 83: IMO if we don't need this `blockquote` element anymore, we should probably just delete the lines.

mattfredericksen (approved):
LGTM! One tiny suggested change.

frontend/components/cards/ModuleUnionCard.tsx
```diff
@@ -8,6 +8,15 @@ interface HighlightEllipsisProps {
   unsafeHTML: string;
 }
 
+function applyHTMLFilter(str: string) {
```
mattfredericksen, line 11: Oh, also maybe rename `string` to `unsafeHTML` (preferred, IMO) or `htmlString`.

PR #771, opened 2021-10-06
MH-256: Add jest and refactor search to facilitate unit testing

mattfredericksen:
Major changes (other than jest tests):
- All pages must now use the extension `.page.tsx`. This is to allow co-locating test files with their respective pages, which otherwise would cause test files to be interpreted as pages.
- `SearchFormContext` was removed in favor of explicitly passing event handlers to each field. This required a fairly extensive refactor of search form fields. The old design was a bad idea to begin with, but it would also have caused challenges for unit testing since the provider would need to be included in all tests.


alashow (commented):
LGTM!

mattfredericksen:
Reviewed by team in https://discord.com/channels/799456464120971294/819742257183850536/896533781916291122

PR #778, opened 2021-10-09
MH-45

aaw0151:


aaw0151:
Old:
image

New:
image

apps/entities/api/schema.py
```diff
@@ -7,12 +7,15 @@
...
 from apps.graph.types import ModuleType
+from apps.quotes.api.schema import QuoteType
 
 
 class EntityType(ModuleType):
     """GraphQL type for the Entity model."""
 
     cached_images = GenericScalar(source='cached_images')
+    related_quotes = graphene.List(QuoteType, slug=graphene.String())
+    # related_entities = graphene.List(CANNOT FIND OWN TYPE, slug=graphene.String())
```
alashow, line 18: Are these commented lines still needed?

alashow (approved):

frontend/pages/entities/[slug].tsx
```diff
@@ -14,11 +28,37 @@ interface EntityProps {
...
+                  )}
+                  <CardContent>
+                    <HTMLEllipsis unsafeHTML={relatedQuote.bite} maxLine="3" basedOn="words" />
+                  </CardContent>
+                </EntityRelatedQuoteCard>
+              </a>
+            </Link>
+          ))}
+      </ModuleContainer>
     </Layout>
```
Comment (mattfredericksen) [inline frontend/pages/entities/[slug].tsx:62]: This can be simplified a little. Currently, we (1) check `hasRelatedQuotes` twice and (2) render an empty `ModuleContainer` even if there are no quotes. I think we should use this structure instead:
```jsx
<Layout>
  <>
    <ModuleContainer/>  // first one
    {entity.relatedQuotes?.length > 0 && (
      <ModuleContainer>
        <CardHeader/>
        {entity.relatedQuotes.map(...)}
      </ModuleContainer>
    }
  </>
</Layout>
```
I'm not 100% sure, but I also think that `ModuleContainer` only provides some margins, so it's possible you can just place the new block right after `ModuleDetail` instead of creating a second `ModuleContainer`.

frontend/pages/entities/[slug].tsx
```diff
@@ -14,11 +28,37 @@ interface EntityProps {
  * A page that renders the HTML of a single entity.
  */
 const EntityDetailPage: FC<EntityProps> = ({ entity }: EntityProps) => {
+  const hasRelatedQuotes = (entity.relatedQuotes && entity.relatedQuotes.length) || undefined;
```
Comment (mattfredericksen) [inline frontend/pages/entities/[slug].tsx:31]: I think we can remove this variable if my other suggestion is accepted.

apps/entities/api/schema.py
```diff
@@ -49,6 +55,10 @@ def resolve_entity(root, info, slug: str) -> Optional[Entity]:
                     pass
         return None
 
+    # @staticmethod
+    # def resolve_related_entities(root, info, slug: str) -> list[Entity]:
+    # return Entity.objects.get(pk=int(slug)).related_entities.all()
+
```
mattfredericksen, line 61: Do we need this comment?

mattfredericksen (changes requested):
Mostly LGTM, just code clarity changes requested.

core/urls.py
```diff
@@ -98,10 +99,10 @@ def error(request: 'HttpRequest'):
     path('healthcheck/', bare_status),  # basic healthcheck
     path('ht/', include('health_check.urls')),
     # ---------------------------------
-    # GraphQL
+    # GraphQL (enabled during DEV, disabled during PROD)
     # ---------------------------------
-    path('graphql/', csrf_exempt(GraphQLView.as_view(graphiql=False))),
-    path('graphiql/', csrf_exempt(GraphQLView.as_view(graphiql=True))),
+    path('graphql/', csrf_exempt(GraphQLView.as_view(graphiql=IS_DEV))),
+    path('graphiql/', csrf_exempt(GraphQLView.as_view(graphiql=IS_DEV))),
```
iacobfred, line 105: Nice.

PR #791, opened 2021-10-16
MH-252: Added "About Us" and "Donation" link

sunzae99:


frontend/components/AboutModularHistory.tsx
```diff
@@ -0,0 +1,23 @@
+import PageHeader from "@/components/PageHeader";
+import { Container, Divider, Grid, Link, Paper } from "@mui/material";
+import React, { FC } from "react";
+
+const AboutModularHistory: FC  = () => {
```
iacobfred, line 5: Hi @sunzae99, since we don't plan on reusing this component in other places, how about defining this component in the page file (frontend/pages/index.tsx) instead of having a separate file for it?

sunzae99:
Here's how it looks currently. Any suggestions for improvements will be highly appreciated.
Capture

frontend/components/AboutModularHistory.tsx
iacobfred: I think we should only use a separate component file if we intend for the component to be reusable. If this component is only displayed on the homepage, I'd prefer to just have it in the homepage file.

PR #814, opened 2021-10-21
MH-248: implement search input in navbar

aaw0151:


aaw0151:
Current view of the search bar implementation:
image
image

mattfredericksen:
@aaw0151 does the responsive behavior still need to be worked out / is any help needed here?

aaw0151:
> @aaw0151 does the responsive behavior still need to be worked out / is any help needed here?

Probably. My update worked on my end as it will simply hide the search bar if the window is too small to prevent the other nav items clipping each other but it's not very elegant. Nor do I know how to make it more elegant as the TextField is very finicky.

PR #822, opened 2021-10-24
MH-253: add subscription section to homepage

aaw0151:
This new component is currently hidden until the subscription section is completely implemented.


frontend/pages/index.page.tsx
```diff
@@ -144,3 +155,89 @@ const AboutModularHistory: FC = () => {
...
+      <Box sx={{ fontWeight: "100" }}>
+        <Typography style={{ fontSize: "10pt" }}>
+          {"By entering your details, you are agreeing to ModularHistory's "}
+          <Link href="/privacy">
+            <a>privacy policy</a>
+          </Link>{" "}
+          and{" "}
+          <Link href="/terms">
+            <a>terms of use</a>
+          </Link>
```
mattfredericksen, line 202: Nit: doesn't need to be changed in this PR, but I think we could use `{" and "}` instead of `{" "}and{" "}`.

mattfredericksen (approved):
LGTM

PR #860, opened 2021-11-02
Docker compose v2 support

alashow:


iacobfred:
@alashow, are these the only two places that `restart_policy` needs to be modified? Could you please add a link to the documentation for this upgrade? Thanks!

alashow:
> @alashow, are these the only two places that `restart_policy` needs to be modified? Could you please add a link to the documentation for this upgrade? Thanks!

Yes, these are the only places needed for me to change to run webserver & django. I'll make sure other containers use updated `restart_policy` too.

https://github.com/compose-spec/compose-spec/blob/master/spec.md#restart

alashow:
Will be force switched via `docker compose` uses in MH-270 https://github.com/ModularHistory/modularhistory/pull/865/commits/b1767a876b272398c6f0e731457d55efce85ed5f

PR #861, opened 2021-11-02
Explicitly set poetry install path

mattfredericksen:
I think it would be preferable to not override poetry's install path, but that would lead to inconsistencies between installation path on Mac and Unix. Any thoughts @iacobfred?

https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py
```py
"""
This script will install Poetry and its dependencies.

It does, in order:

  - Downloads the virtualenv package to a temporary directory and add it to sys.path.
  - Creates a virtual environment in the correct OS data dir which will be
      - `%APPDATA%\\pypoetry` on Windows
      -  ~/Library/Application Support/pypoetry on MacOS
      - `${XDG_DATA_HOME}/pypoetry` (or `~/.local/share/pypoetry` if it's not set) on UNIX systems
      - In `${POETRY_HOME}` if it's set.
  - Installs the latest or given version of Poetry inside this virtual environment.
  - Installs a `poetry` script in the Python user directory (or `${POETRY_HOME/bin}` if `POETRY_HOME` is set).
"""
```

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -306,14 +306,15 @@ if [[ ! "$(pyenv versions)" =~ "$PYTHON_VERSION" ]]; then
 fi
 
 # Install and configure Poetry.
+export POETRY_HOME="$HOME/.poetry"
 # shellcheck disable=SC2016
 poetry_init='export PATH="$HOME/.poetry/bin:$PATH"'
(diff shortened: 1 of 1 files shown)


iacobfred (approved):
LGTM

PR #862, opened 2021-11-03
Mh 268 Dynamic forms

aaw0151:
Dynamic form for content creation.


aaw0151:
image
image
image
image
image
image
image
image

aaw0151:
Currently, this component is displaying in the homepage. This will be removed before merging.

frontend/pages/index.page.tsx
```diff
@@ -1,3 +1,4 @@
+import DyanmicForm from "@/components/forms/DynamicForm";
```
iacobfred, line 1: typo: `Dyanmic` -> `Dynamic`

frontend/components/forms/DynamicForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+  InputLabel,
+  MenuItem,
+  OutlinedInput,
+  Select,
+} from "@mui/material";
+import Card from "@mui/material/Card";
+import TextField from "@mui/material/TextField";
+import { FC, useEffect, useState } from "react";
+
+const DyanmicForm: FC = () => {
```
iacobfred, line 17: typo: `Dyanmic` -> `Dynamic`

frontend/pages/index.page.tsx
aaw0151: How embarrassing. Fixed!

mattfredericksen: Not embarassing :)

frontend/components/forms/DynamicForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+const fields = ["topics"] as const;
+type Field = typeof fields[number];
+type FieldsRef = MutableRefObject<Record<Field, any>>;
+type FieldCallbacks = Record<Field, (value: ParsedUrlQueryInput[string]) => void>;
+
+//Dynamic fields form
+//Can be called using the DynamicForm component or by using the DynamicFormFields prop and specifying the 'type' prop
+const DynamicFormFields: FC<DynamicFormProps> = ({ type }: DynamicFormProps) => {
+  const [formData, setFormData] = useState([]);
+  const [choiceValue, setChoiceValue] = useState<string[]>([]);
```
mattfredericksen, line 85: It looks like the same choice state values here are used for all `Select` fields rendered, so it breaks when more than one value is selected.

image

frontend/components/forms/DynamicForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const isLoading = useContext(PageTransitionContext);
+
+  const handleChoice = (event: { target: { value: string[] | string } }) => {
+    const value = event.target.value as string[] | string;
+    setChoiceValue(typeof value === "string" ? value.split(",") : value);
+  };
+
+  const fieldsRef = useRef(
+    Object.fromEntries(fields.map((name) => [name, router.query[name]]))
+  ) as FieldsRef;
```
mattfredericksen, line 103: Do we want/need to initialize field values from URL parameters? At least right now they are all undefined, so we may as well initialize the useRef with {}.

mattfredericksen (changes requested):
Looking good so far, but needs some bug fixes.

frontend/pages/entities/[slug].page.tsx
```diff
@@ -38,7 +38,7 @@ const EntityDetailPage: FC<EntityProps> = ({ entity }: EntityProps) => {
       />
       <ModuleContainer>
         <ModuleDetail module={entity} />
-        {entity.relatedQuotes && entity.relatedQuotes.length && (
+        {entity.relatedQuotes && entity.relatedQuotes.length > 0 && (
```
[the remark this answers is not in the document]
Comment (aaw0151) [reply frontend/pages/entities/[slug].page.tsx]: For some reason, a stray "0" will appear on the page if the entity has no related quotes. Using `> 0` fixes this.

frontend/components/forms/DynamicForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+    return obj.editable;
+  };
+
+  const createDisplayName = (str: string) => {
+    return str.replace(/_/g, " ");
+  };
+
+  const createChoiceValue = (str: string) => {
+    return str.replace(/<.*?>/g, "");
+  };
```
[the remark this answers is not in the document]
aaw0151: It was on the `related_quotes` field for entities IIRC.

frontend/pages/entities/[slug].page.tsx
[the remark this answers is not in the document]
Comment (mattfredericksen) [reply frontend/pages/entities/[slug].page.tsx]: Do you mean using conditional chaining can still return a `0`? When I test it, the output is always `false`.

frontend/components/forms/DynamicForm.tsx
aaw0151: Hmm okay. Would I need to create a state value for each ManyRelatedField that is displayed before hand? Or is there an easier way of doing this?

frontend/pages/entities/[slug].page.tsx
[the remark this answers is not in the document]
Comment (aaw0151) [reply frontend/pages/entities/[slug].page.tsx]: When using `...relatedQuotes.length && (...` there will be a "0" that appears on the page. When using `...relatedQuotes.length > 0 && (...` the "0" goes away. I made some screenshots of the behavior here: https://github.com/ModularHistory/modularhistory/pull/842#issuecomment-953506370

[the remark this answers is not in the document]
Comment (mattfredericksen) [reply frontend/pages/entities/[slug].page.tsx]: Oh, yeah that's fine, my suggestion is just to use optional chaining (still including `> 0`).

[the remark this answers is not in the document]
Comment (aaw0151) [reply frontend/pages/entities/[slug].page.tsx]: Oh my mistake, I misread your first comment, will fix now.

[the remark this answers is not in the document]
Comment (aaw0151) [reply frontend/pages/entities/[slug].page.tsx]: image
I remember why I added the excessive check for the `entity`, TypeScript kept complaining about possibly undefined variables.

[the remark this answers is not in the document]
Comment (mattfredericksen) [reply frontend/pages/entities/[slug].page.tsx]: Huh, I'm surprised I haven't come across this before. It looks like we can fix this by using 
```jsx
{(entity.relatedQuotes?.length ?? 0) > 0 && (
```
which is still better than repeating `entity.relatedQuotes` IMO.

frontend/components/forms/DynamicForm.tsx
mattfredericksen: https://github.com/ModularHistory/modularhistory/blob/e9ee643744f370a0c4a4fcdd1b1de8f02d295782/frontend/components/forms/DynamicForm.tsx#L99-L102

The way this code works is to have a single object `fieldsRef` holding values for each field, and have unique callbacks for each field so that each write to that object with a unique key. You might just be able to pass the correct callback `fieldCallbacks.current[field.name]` to the `Select` components, but it's also possible the generated callback would need to be modified to handle `Select`'s `onChange` arguments ([docs](https://mui.com/api/select/#props)). If that's true, then you could just add a field type check to L100-102, something like
```jsx
fields.map(({name, type}) => (
  [
    name, 
    type === "M2M" 
      ? (whatever, args, value) => {fieldsRef.current[name] = value} 
      : (value) => {fieldsRef.current[name] = value} 
  ]
))
```

It looks like you're hardcoding the fields right now instead of pulling names from the API response, so that will need to be fixed. We can only do that on the search form since the fields are hardcoded.
https://github.com/ModularHistory/modularhistory/blob/e9ee643744f370a0c4a4fcdd1b1de8f02d295782/frontend/components/forms/DynamicForm.tsx#L75-L76

aaw0151: Currently, in this implementation, Jacob said that it would be a good idea to just copy over the topics search from the SearchForm.tsx. This is currently the only field name that is hardcoded as that topic field use the topics search instead of creating its own field.

This isn't the best as it is hardcoded but currently, all types have a topic field in the API response.

mattfredericksen: As far as I can tell, utilizing the pre-existing topics field is handled entirely by the logic here.
https://github.com/ModularHistory/modularhistory/blob/e9ee643744f370a0c4a4fcdd1b1de8f02d295782/frontend/components/forms/DynamicForm.tsx#L130-L132
It doesn't look like it depends on having this array hard-coded.

alashow (approved):
LGTM

iacobfred (approved):
Let's merge this! There are some things to fix, but I don't want this to slow down other work in the meantime.

mattfredericksen:
@aaw0151 can you create a follow-up ticket for fixing this?
https://github.com/ModularHistory/modularhistory/pull/862#discussion_r744380099

PR #864, opened 2021-11-05
Homepage style improvements

iacobfred:


frontend/pages/index.page.tsx
```diff
@@ -191,53 +172,35 @@ const FeaturedContent: FC = () => {
...
-                ) : (
-                  <p>Sorry, there are no featured contents to display. Please check back later!</p>
-                )}
-              </>
-            </Grid>
-          </Box>
-        </Card>
+            ) : (
+              <p>Sorry, there are no featured contents to display. Please check back later!</p>
+            )}
```
mattfredericksen, line 203: Wasn't introduced in this PR, but can we change this to "There is no featured content ..."?

frontend/pages/index.page.tsx
```diff
@@ -292,37 +255,27 @@ const SubscriptionBox: FC = () => {
...
-              <CloseIcon />
-            </Button>
-          </Grid>
-          <Typography>Sign up to recieve our newsletter!</Typography>
-          {subscriptionForm}
-          {subscribeDisclaimer}
-        </CardContent>
-      </Card>
-    </Box>
+    <StyledCard elevation={5}>
```
mattfredericksen, line 262: This isn't a problem necessarily, but we should consider using the `raised` prop instead of `elevation={magicNumber}` if it looks similar enough.

frontend/pages/index.page.tsx
```diff
@@ -191,53 +172,35 @@ const FeaturedContent: FC = () => {
...
-                <Typography variant="h6" gutterBottom component="div" fontWeight="bold">
-                  Featured Content
-                </Typography>
+    <Box>
+      <Box sx={{ m: 3, alignItems: "center", justifyContent: "center" }}>
+        <Grid container alignItems="center" justifyContent="center">
+          <>
+            {items.length ? (
+              items.map((module, index) => (
+                <Grid item key={index} xs={12} sm={6} md={4} lg={3} xl={3}>
```
mattfredericksen, line 185: Breakpoints propagate up, so we can remove `xl`.

frontend/pages/index.page.tsx
```diff
@@ -59,111 +89,68 @@ export default function Home() {
...
-                <Link href="/about/contributing" variant="body2">
-                  {"contribute content"}
-                </Link>{" "}
-                and/or{" "}
-                <Link href="/donations" variant="body2">
-                  {" donate."}
-                </Link>
-              </p>
-            </Grid>
+    <Box>
```
mattfredericksen, line 132: Nit: `<Box>` &rarr; `<>`. Not sure if it's better for performance, but it's better idiomatically.

frontend/pages/index.page.tsx
```diff
@@ -191,53 +178,35 @@ const FeaturedContent: FC = () => {
...
-        }}
-      >
-        <Card elevation={5}>
-          <Box sx={{ m: 3, alignItems: "center", justifyContent: "center" }}>
-            <Grid container alignItems="center" justifyContent="center">
-              <Grid item>
-                <Typography variant="h6" gutterBottom component="div" fontWeight="bold">
-                  Featured Content
-                </Typography>
+    <Box>
```
mattfredericksen, line 179: I think this outer `<Box/>` can be removed.

frontend/pages/index.page.tsx
```diff
@@ -191,53 +178,35 @@ const FeaturedContent: FC = () => {
...
+                <Grid item key={index} xs={12} sm={6} md={4} lg={3} xl={3}>
+                  <Link href={module.absoluteUrl} underline="none">
+                    <a>
+                      <ModuleUnionCard module={module} />
+                    </a>
+                  </Link>
+                </Grid>
+              ))
+            ) : loading ? (
+              <Grid item xs={12} sm={6} md={4} lg={3} xl={3}>
```
mattfredericksen, line 194: Same `xl` thing.

frontend/components/TodayInHistory.tsx
```diff
@@ -30,37 +29,32 @@ const TodayInHistory: FC = () => {
...
+    <Grid container spacing={1} justifyContent="center" alignItems="center">
+      <>
+        {items.length ? (
+          items.map((module, index) => (
+            <Grid item key={index}>
+              <Link href={module.absoluteUrl}>
+                <a>
+                  <ModuleUnionCard module={module} />
+                </a>
+              </Link>
```
mattfredericksen, line 41: Not related to this PR, but we use the pattern enough that we should just always render `Link` inside `ModuleCard`, and maybe add a prop if we need to disable it.

frontend/pages/index.page.tsx
```diff
@@ -59,111 +89,66 @@ export default function Home() {
...
-        </Grid>
-      </Grid>
-      <Grid container spacing={2} sx={{ p: 4 }}>
-        <Grid item xs={12}>
-          <Card elevation={5}>
+          </StyledCard>
+        </GridItem>
+        <GridItem>
+          <StyledCard elevation={5}>
+            <StyledCardHeader title="Module Search" />
```
mattfredericksen, line 101: This is what `raised` looks like (only applied to "Module Search" card).

image

mattfredericksen (approved):
LGTM, just code quality nits.

frontend/pages/index.page.tsx
```diff
@@ -59,111 +89,66 @@ export default function Home() {
...
-        <Grid item xs={12} sm={6} md={6}>
-          <Container>
-            <FeaturedContent />
-          </Container>
-        </Grid>
-      </Grid>
-      <Grid container spacing={2} sx={{ p: 4 }}>
-        <Grid item xs={12}>
-          <Card elevation={5}>
+          </StyledCard>
```
mattfredericksen, line 97: `<AboutModularHistory />` should be wrapped in `CardContent`.

frontend/pages/index.page.tsx
```diff
@@ -59,111 +89,66 @@ export default function Home() {
...
+        <GridItem>
+          <StyledCard elevation={5} sx={{ minWidth: "20rem" }}>
+            <StyledCardHeader title="Today in History" />
+            <TodayInHistory />
+          </StyledCard>
+        </GridItem>
+        <GridItem>
+          <StyledCard elevation={5} hidden>
+            <SubscriptionBox />
+          </StyledCard>
```
mattfredericksen, line 123: Likewise for `<FeaturedContent />`, `<TodayInHistory />`, `<SubscriptionBox />`. Seems like we can safely remove the styling on lines like 180, which were only needed because `CardContent` was missing.
https://github.com/ModularHistory/modularhistory/blob/c4f45f2c0ee06c82398330d7f6a554f548a1f855/frontend/pages/index.page.tsx#L179-L181

mattfredericksen (changes requested):

mattfredericksen (approved):
LFGTM!

PR #865, opened 2021-11-06
MH-270: moderation API tests

iacobfred:


apps/moderation/api/tests.py
```diff
@@ -0,0 +1,205 @@
...
+        object_id: Optional[int] = None,
+    ):
+        if url_kwargs is None:
+            url_kwargs = {}
+        if object_id:
+            url_kwargs.update({'pk_or_slug': object_id})
+        path = reverse(f'{self.api_name}:{view}', kwargs=url_kwargs)
+        # Force request path to include API path suffix.
+        if self.api_path_suffix and self.api_path_suffix not in path:
+            path += self.api_path_suffix + '/'
```
alashow, line 81: We might need to add this suffix for `_test_api_view_get` too. I'm guessing it's currently not failing because it's probably accessing parent path, i.e for articles, `/api/sources/` instead of `/api/sources/articles/`

apps/dates/__init__.py
```diff
@@ -0,0 +1 @@
+from .fakers import HistoricDateTimeProvider
```
iacobfred, line 1: Looks like this breaks building the Django image outside of dev environment, since `factory-boy` is a dev-only dependency

alashow (approved):

apps/sources/api/sources/article/tests.py
```diff
@@ -0,0 +1,95 @@
...
+            'end_date': '2020-01-01 01:01:20.086202',
+            'publication': PublicationDrfSerializer(self.publications[0]).data,
+            'attributions': [
+                SourceAttributionDrfSerializer(
+                    Article.attributees.through(
+                        attributee_id=id,
+                    )
+                ).data
+                for id in self.attributee_ids[:2]
+            ],
```
alashow, line 70: Can't we use normal relation fields for publication and attribution/attributees? Looks like tags are working

Otherwise it complicates

alashow (approved):

apps/sources/api/sources/article/tests.py
iacobfred: I think it depends... For reading, I think we want to use nested serializers. But for writing, maybe we don't need to use nested serializers. Maybe one solution could be to use, e.g., a read-only `publication` field (using serializer) and a write-only `publication_id` field. Relation fields like `attribution` are another problem; I'll respond about that separately.

alashow:
We already have a solution for publication: read only version with
_serialized suffix, write-only version for id (similar to Speech.utterrance)

PR #867, opened 2021-11-07
MH-74: Allow user to share collections

sunzae99:


frontend/pages/collections/[slug].page.tsx
```diff
@@ -78,3 +103,113 @@ export const getStaticPaths: GetStaticPaths = async () => {
...
+    </>
+  );
+};
+
+const DesktopViewComponent: FC<CollectionProps> = ({ collection }: CollectionProps) => {
+  return (
+    <Container>
+      <Grid sx={{ m: 3, display: "flex", alignItems: "center", justifyContent: "flex-end" }}>
+        <Grid sx={{ maxWidth: 250, alignItems: "center", justifyContent: "center" }}>
+          <Grid container spacing={2} alignItems="center" justifyContent="center">
```
Comment (mattfredericksen) [inline frontend/pages/collections/[slug].page.tsx:185]: This level of nested components is almost certainly unnecessary. @sunzae99 can you experiment with removing props/grids and see if this can be reduced?

frontend/pages/collections/[slug].page.tsx
```diff
@@ -78,3 +103,113 @@ export const getStaticPaths: GetStaticPaths = async () => {
...
   };
 };
+
+const MobileViewComponent: FC<CollectionProps> = ({ collection }: CollectionProps) => {
+  const [showIcons, setShowIcons] = useState(false);
+  const onClick = () => {
+    setShowIcons(true);
+    if (showIcons) {
+      setShowIcons(false);
+    }
```
Comment (mattfredericksen) [inline frontend/pages/collections/[slug].page.tsx:113]: `setState` can also take a function that has the previous state as the parameter, so this can be simplified to 
```jsx
setShowIcons((prevState) => !prevState)
```

frontend/pages/collections/[slug].page.tsx
```diff
@@ -78,3 +103,113 @@ export const getStaticPaths: GetStaticPaths = async () => {
...
+        <Grid sx={{ m: 3, display: "flex", alignItems: "center", justifyContent: "flex-end" }}>
+          <Grid sx={{ maxWidth: 250, alignItems: "center", justifyContent: "center" }}>
+            <Grid
+              sx={{
+                flexDirection: "column",
+                display: "flex",
+                alignItems: "center",
+                justifyContent: "center",
+              }}
+            >
```
Comment (mattfredericksen) [inline frontend/pages/collections/[slug].page.tsx:127]: Same as my previous comment about nesting.

frontend/pages/collections/[slug].page.tsx
```diff
@@ -78,3 +103,113 @@ export const getStaticPaths: GetStaticPaths = async () => {
...
+                      </TwitterShareButton>
+                    </Grid>
+                    <Grid item>
+                      <EmailShareButton
+                        url={`https://modularhistory.com/collections/${collection.slug}`}
+                        subject={collection.title}
+                      >
+                        <EmailIcon size={36} round={true} />
+                      </EmailShareButton>
+                    </Grid>
```
Comment (mattfredericksen) [inline frontend/pages/collections/[slug].page.tsx:168]: Looks like this code is duplicated between mobile and desktop view components. It can be moved to a common child component.

frontend/pages/collections/[slug].page.tsx
```diff
@@ -78,3 +103,113 @@ export const getStaticPaths: GetStaticPaths = async () => {
...
+const MobileViewComponent: FC<CollectionProps> = ({ collection }: CollectionProps) => {
+  const [showIcons, setShowIcons] = useState(false);
+  const onClick = () => {
+    setShowIcons(true);
+    if (showIcons) {
+      setShowIcons(false);
+    }
+  };
+  return (
+    <>
```
Comment (mattfredericksen) [inline frontend/pages/collections/[slug].page.tsx:116]: Looks like this wrapping fragment is unnecessary.

frontend/pages/collections/[slug].page.tsx
```diff
@@ -78,3 +103,113 @@ export const getStaticPaths: GetStaticPaths = async () => {
     fallback: "blocking",
   };
 };
+
+const MobileViewComponent: FC<CollectionProps> = ({ collection }: CollectionProps) => {
```
Comment (mattfredericksen) [inline frontend/pages/collections/[slug].page.tsx:107]: These component names are a little generic. Can we use more descriptive names? Maybe `MobileSharingButtonGroup` or `MobileSharingButtons`?

mattfredericksen (changes requested):

frontend/pages/collections/[slug].page.tsx
Comment (iacobfred) [reply frontend/pages/collections/[slug].page.tsx:185]: +1, let's use just a single `<Grid container ... >` to wrap these grid items.

We also shouldn't need the `<Container>`. If it adds padding to the grid, then if necessary, we can just add the padding to the grid `sx` instead.

Comment (mattfredericksen) [reply frontend/pages/collections/[slug].page.tsx:185]: > we can just add the padding to the grid via `sx` instead.

We actually don't even need `sx`. https://mui.com/api/grid/:
> As a CSS utility, the Grid component also supports all system properties. You can use them as props directly on the component.

iacobfred (approved):

frontend/pages/collections/[slug].page.tsx
Comment (sunzae99) [reply frontend/pages/collections/[slug].page.tsx]: Resolved in a new commit!

frontend/pages/collections/[slug].page.tsx
```diff
@@ -78,3 +103,71 @@ export const getStaticPaths: GetStaticPaths = async () => {
...
+          </TwitterShareButton>
+        </Grid>
+        <Grid item>
+          <EmailShareButton
+            url={`https://modularhistory.com/collections/${collection.slug}`}
+            subject={collection.title}
+          >
+            <EmailIcon size={36} round={true} />
+          </EmailShareButton>
+        </Grid>
```
Comment (mattfredericksen) [inline frontend/pages/collections/[slug].page.tsx:169]: Common props should be extracted to a separate object and then spread on each component. There's too much duplication here. Duplication reduces maintainability.
E.g.
```tsx
const url = `https://modularhistory.com${collection.absoluteUrl}`;
const iconProps = { size: 36, round: true };
// ...
    <FacebookShareButton
      url={url}
      quote={collection.title}
    >
      <FacebookIcon { ...iconProps } />
    </FacebookShareButton>
// etc
```

mattfredericksen (changes requested):
LGTM, just one change.

frontend/pages/collections/[slug].page.tsx
Comment (mattfredericksen) [reply frontend/pages/collections/[slug].page.tsx:169]: @iacobfred I see that the `BASE_URL` env var is set to `http://modularhistory.dev.net`, is it `https://modularhistory.com` in production? If so we can add that var to the exported `env` here
https://github.com/ModularHistory/modularhistory/blob/d04247ee7b8127390cc81629fd7a0896d37b137d/frontend/next.config.js#L76-L80
which would let us use `process.env.BASE_URL` instead of the static string.

mattfredericksen (approved):
LGTM

PR #878, opened 2021-11-11
MH-271: Improve Fields API for relation fields & choices for typed modules + Source models moderated API tests

alashow:
- [x] Add choices for typed modules
- [x] Disable listing choices for one2one/m2m fields like `tags`, `related_entities`, `publication`, `images` and add `instant_search` field to hint api clients what instant_search endpoint to use and possible filters needed.


mattfredericksen:
Converted to draft, sorry if that's unwanted.

mattfredericksen (commented):
@alashow when I run `docker-compose exec django pytest` or `poetry run invoke test`, it seems like most of the tests fail. Does this happen for you also or am I running the tests incorrectly?

alashow:
@mattfredericksen  I am running them using `poetry run pytest` (takes ~60 seconds).

`docker-compose exec django pytest` took 11 minutes (possibly due to docker restrictions)
`poetry run invoke test` failed with weird errors for me too.

mattfredericksen (commented):
Looks like randomly generated data is sometimes too long?
image

This happened once after running test ~5 times 🤷‍♂️

alashow:
> Looks like randomly generated data is sometimes too long? 
> 
> This happened once after running test ~5 times 🤷‍♂️

Yeah I had to tweak factory config in some other places too.

mattfredericksen (commented):
Sorry for the piecemeal review.

Not sure if this should be addressed by this PR, but it looks like some fields can be `required=True` and `editable=False`. I'm guessing the field is required on the model, but it autogenerated so cannot be set by the contributor. I guess this won't cause any problems if we always exclude fields on the frontend with `editable=False`?
image

I would expect this field on `Entity` to use an ES endpoint instead of a choices array. Should this be changed? (Unfortunate that `Birth` is identical to `Occurrence` currently, so it seems like we can't filter `type="proposition.occurrence.birth"` or something).
image

alashow:
Yeah it should be to exclude non editable fields, we might do it on the
backend.

I forgot to add instant search fields for birth/death. Although we can
re-use propositions InstantSearch index for this, like in Speech.utterance
https://github.com/ModularHistory/modularhistory/blob/7ac0eee8d5759c5ac84f789e87dac8a805c6559f/apps/sources/api/sources/speech/serializers.py#L11-L18

mattfredericksen (approved):
LGTM

docker-compose.yml
```diff
@@ -226,8 +226,8 @@ services:
       - elasticsearch
     env_file: .env
     environment:
-      ELASTICSEARCH_URL: http://elasticsearch:9200
-      ELASTICSEARCH_HOSTS: http://elasticsearch:9200
+      ELASTICSEARCH_URL: https://elasticsearch:9200
```
iacobfred, line 229: Interesting, why?

apps/moderation/serializers.py
```diff
@@ -1,22 +1,66 @@
...
+                    'required': not getattr(field, 'required', False),
+                    'allow_blank': not getattr(field, 'allow_blank', False),
+                    'verbose_name': getattr(field, 'verbose_name', None),
+                    'help_text': getattr(field, 'help_text', None),
+                    'choices': self.get_choices_for_field(field, field_name),
+                }
+                instant_search_field = self.get_instant_search_field(field_name)
+                if instant_search_field:
+                    data.update(data | {'instant_search': instant_search_field})
+                fields.append(data)
```
mattfredericksen, line 59: I think there's a redundant merge here. Maybe we could do 
```py
data.update({'instant_search': instant_search_field})
```
or
```py
data['instant_search'] = instant_search_field
```

mattfredericksen (commented):
Nit

apps/moderation/serializers.py
```diff
@@ -1,22 +1,66 @@
...
 
-    class ModeratedModelSerializer(model_serializer):
-        fields = serializers.ReadOnlyField(source='get_moderated_fields')
+class ModeratedModelSerializer(serializers.ModelSerializer):
+    moderated_fields_excludes = ['id', 'meta', 'admin_url', 'absolute_url', 'cached_tags']
+
+    instant_search_fields = {}
+
+    def get_instant_search_fields(self) -> dict:
+        return common_instant_search_fields | self.instant_search_fields
```
mattfredericksen, line 19: What is the rationale behind `common_instant_search_fields`? Could they not be defined on their corresponding serializers, or does it complicate inheritance?

apps/moderation/serializers.py
```diff
@@ -1,22 +1,66 @@
...
+
```
mattfredericksen, line 20: Not sure how much this matters, but could we either improve the efficiency of checking instant search fields, or decorate this method with [`cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property)? Seems like it is currently being called for each field.

apps/moderation/serializers.py
alashow: As the name suggests, these m2m fields are used in many modules. Extending serializers use instant_search_fields to define their own.

apps/moderation/serializers.py
alashow: I will look into usage of cached_property, but it shouldn't matter much. It's only called when building fields for Fields API, and since it's only merging dict's, it's fast enough.

docker-compose.yml
alashow: I think Kibana was failing with ssl errors until I did this.

apps/moderation/serializers.py
mattfredericksen: Oh, right. I was assuming it was being called for each request/serialization.

apps/images/factories.py
```diff
@@ -28,9 +28,9 @@ class Meta:
     image = ImageField(from_func=generate_temporary_image)
     width = fuzzy.FuzzyInteger(500, 1000)
     height = fuzzy.FuzzyInteger(500, 1000)
-    title = factory.Faker('sentence', nb_words=10)
+    title = factory.Faker('sentence', nb_words=3)
```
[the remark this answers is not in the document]
alashow: It was making the tests flaky because sometimes 10 words length would be larger than allowed max length.

https://github.com/ModularHistory/modularhistory/pull/878/commits/7ac0eee8d5759c5ac84f789e87dac8a805c6559f

apps/moderation/serializers.py
alashow: Second one night work the same, but first one seems like it would replace `data` with just given dict?

mattfredericksen: I think it only replaces overlapping keys, which is the same behavior as the `|` operator.

alashow: Gotcha, for some reason I assumed `update` would clear & set.

iacobfred (approved):
LGTM!

docker-compose.yml
alashow: I'm gonna revert this. I'm not sure why but it's not working on my machine either way (`http` or `https`). It starts up, but I can't access it at `localhost:5601`

PR #884, opened 2021-11-15
Mh 91 Alt text to images

aaw0151:


aaw0151:
Before update:
image

After update:
image

aaw0151:
I only changed one file `/modularhistory/frontend/components/cards/ModuleCard.tsx`. I'm not sure why my files from https://github.com/ModularHistory/modularhistory/pull/862 are being committed as well.

@mattfredericksen @iacobfred do you know why that would be?

mattfredericksen:
> I only changed one file `/modularhistory/frontend/components/cards/ModuleCard.tsx`. I'm not sure why my files from #862 are being committed as well.

I think that just means at you branched from `MH-268` instead of `main` when creating this branch. I'll see if I can correct it.

aaw0151:
> > I only changed one file `/modularhistory/frontend/components/cards/ModuleCard.tsx`. I'm not sure why my files from #862 are being committed as well.
> 
> I think that just means at you branched from `MH-268` instead of `main` when creating this branch. I'll see if I can correct it.

Ah I probably did do that, my bad.

frontend/components/cards/ModuleCard.tsx
```diff
@@ -143,9 +143,14 @@ const ModuleCard: FC<ModuleCardProps> = ({
   let bgImage: Image | undefined;
   if (module.model == "images.image") {
     bgImage = module as Image;
+    //bgImage.title = module.title;
```
mattfredericksen, line 146: If this line is unneeded it should be removed.

mattfredericksen, line 146: If this line is unneeded it should be removed.

mattfredericksen (commented):
@iacobfred leaving review to you since this affects site content.

iacobfred (commented):
Hmmm, it looks like this PR currently does not add an alt attribute to images. It is adding an `aria-label` attribute to the cards that have image backgrounds, which I don't think is necessary, since screen readers are able to read the textual content of these cards. We probably don't want screen readers to read the caption of the background image for each result on the SERP; we just want the screen readers to read the card titles and truncated module descriptions.

We specifically want `alt` attributes on `img` elements because pages can be penalized by Google if they are missing `alt` text for images. Elements with image backgrounds are fine, though, because the background is just there for style and doesn't necessarily need to be communicated to blind people.

Another thing to consider: since we use `figcaption` elements to display captions underneath our images (including images that are rendered with `img` tags rather than as div backgrounds), and since screen readers will read both the `alt` and `figcaption` content, we should differentiate the two. I just read this article about the difference in usage of `alt` and `figcaption`: https://thoughtbot.com/blog/alt-vs-figcaption

Basically, for an `img` element that depicts an image of Mark Twain sitting in a chair in 1875, the `alt` text would be something like "A man sitting in a chair" and the ficaption text would be something like "Mark Twain, 1875."

So, in order to complete MH-91, I think we should do the following:
- Add an `alt_text` field to the Image model
- Search for all instances of `{% if image.alt_text %}{{ image.alt_text }}{% endif %}` (which will render `alt=""` if there's no alt text value).

iacobfred (changes requested):
See https://github.com/ModularHistory/modularhistory/pull/884#pullrequestreview-810777224

PR #885, opened 2021-11-16
MH-283: Server-side render the TodayInHistory and FeaturedContent components

sunzae99:


frontend/pages/index.page.tsx
```diff
@@ -281,3 +262,31 @@ const SubscriptionBox: FC = () => {
...
+    });
+
+  await axios
+    .get(`http://django:8000/api/home/features/`)
+    .then((response) => {
+      featuredcontentData = response.data;
+    })
+    .catch((error) => {
+      console.error(error);
+    });
```
mattfredericksen, line 286: By having these sequential `await`s, we don't send the second request until the first has completed. We can improve performance by sending both requests before waiting on them.
```ts
    const tihPromise = axios.get(`http://django:8000/api/home/today_in_history/`)
    // ...
    const featuredPromise = axios.get(`http://django:8000/api/home/features/`)
    // ...
    await Promise.allSettled([tihPromise, featuredPromise]);
```

frontend/pages/index.page.tsx
```diff
@@ -157,28 +164,10 @@ const AboutModularHistory: FC = () => {
...
-    };
-  }, [loading]);
-
+const FeaturedContent: FC<FeaturedContentProps> = ({
+  featuredcontentData,
+}: FeaturedContentProps) => {
+  const items = featuredcontentData || [];
   return (
     <Grid container alignItems="center" justifyContent="center">
       <>
```
mattfredericksen, line 173: This fragment doesn't seem to be doing anything.

frontend/pages/index.page.tsx
```diff
@@ -281,3 +262,31 @@ const SubscriptionBox: FC = () => {
     </StyledCard>
   );
 };
+
+export const getStaticProps: GetStaticProps = async () => {
+  let todayinhistoryData = {};
+  let featuredcontentData = {};
```
mattfredericksen, line 268: These should start as empty arrays, since they are arrays when set by the api calls.

frontend/pages/index.page.tsx
```diff
@@ -157,28 +164,10 @@ const AboutModularHistory: FC = () => {
...
-      });
-    return () => {
-      cancelTokenSource.cancel("component unmounted");
-    };
-  }, [loading]);
-
+const FeaturedContent: FC<FeaturedContentProps> = ({
+  featuredcontentData,
+}: FeaturedContentProps) => {
+  const items = featuredcontentData || [];
```
mattfredericksen, line 170: If we default to `[]` instead of `{}` in `getStaticProps`, we won't need this check here and we can just always reference `featuredcontentData` instead of `items`.

frontend/pages/index.page.tsx
```diff
@@ -38,13 +38,19 @@ interface GridItemProps {
   sx?: SxProps;
 }
 
+interface FeaturedContentProps {
+  featuredcontentData: Exclude<ModuleUnion, Topic>[];
+}
```
mattfredericksen, line 43: I think a prop name like `featuredModules` might be better here. I dislike the use of "data," since it doesn't seem to convey any helpful information. The same could be said about `todayinhistoryData`, maybe `todayModules` for that one? It's not great, but better than `todayinhistoryData` IMO.

mattfredericksen (changes requested):

mattfredericksen:
Jest tests for TIH need to be updated.

frontend/components/TodayInHistory.tsx
```diff
@@ -1,33 +1,15 @@
...
-        if (axios.isCancel(error)) return;
-        console.error(error);
-        setLoading(false);
-      });
-    return () => {
-      cancelTokenSource.cancel("component unmounted");
-    };
-  }, [loading]);
+export interface TodayInHistoryProps {
+  todayinhistoryData: Exclude<ModuleUnion, Topic>[];
```
iacobfred, line 8: I'd like to rename this from `todayinhistoryData` to `items`.

Rationale:
- `todayinhistory` part is redundant since this is a param name for the `TodayInHistory` component
- Since it's an array, a plural variable name like `items` is more appropriate than `data` (which usually implies an object with keys)

frontend/pages/index.page.tsx
iacobfred: Yup, `{}` --> `[]`.

iacobfred (changes requested):

frontend/pages/index.page.tsx
sunzae99: Resolved: featuredcontentData -> featuredModules and todayinhistoryData -> todayModules

frontend/components/TodayInHistory.tsx
sunzae99: Changed todayinhistoryData -> todayModules. Is it better to stick to this one or use "items"?

iacobfred (approved):

frontend/components/__tests__/TodayInHistory.test.tsx
```diff
@@ -27,85 +27,38 @@ describe("TodayInHistory", () => {
...
-      const { findByText } = render(<TodayInHistory />);
-      expect(
-        await findByText("There are no modules associated with this date.")
-      ).toBeInTheDocument();
+  describe("module card clicked", () => {
+    it("renders clicked module card", async () => {
+      mockRouter.setCurrentUrl("/");
+      const { findByText } = render(<TodayInHistory modules={[testModule]} />);
+      userEvent.click(await findByText(testModule.title));
+      expect(Router.asPath).toStrictEqual(testModule.absoluteUrl);
```
mattfredericksen, line 61: This test looks the exact same as https://github.com/ModularHistory/modularhistory/blob/07eb6d7a1c2cb9735622ce30e396359e713f75fb/frontend/components/__tests__/TodayInHistory.test.tsx#L39-L44

I think we don't need it.

frontend/components/TodayInHistory.tsx
```diff
@@ -1,59 +1,29 @@
...
-        if (axios.isCancel(error)) return;
-        console.error(error);
-        setLoading(false);
-      });
-    return () => {
-      cancelTokenSource.cancel("component unmounted");
-    };
-  }, [loading]);
+export interface TodayInHistoryProps {
+  modules: Exclude<ModuleUnion, Topic>[];
```
mattfredericksen, line 8: We can replace every usage of `Exclude<ModuleUnion, Topic>` with `SerpModule`.
https://github.com/ModularHistory/modularhistory/blob/07eb6d7a1c2cb9735622ce30e396359e713f75fb/frontend/types/modules.d.ts#L136

mattfredericksen (changes requested):
LGTM, just a couple nits (and [Jacob's comment](https://github.com/ModularHistory/modularhistory/pull/885/files#r752870989)).

frontend/components/__tests__/TodayInHistory.test.tsx
sunzae99: Removed it!

frontend/components/__tests__/TodayInHistory.test.tsx
```diff
@@ -2,8 +2,8 @@ import TodayInHistory from "@/components/TodayInHistory";
 import { Quote } from "@/types/modules";
 import { render } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { rest } from "msw";
-import { setupServer } from "msw/node";
+// import { rest } from "msw";
```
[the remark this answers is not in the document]
sunzae99: Removed :)

iacobfred (approved):

PR #886, opened 2021-11-18
MH-269: Added related entities to entity pages.

aaw0151:


aaw0151:
Currently a work in progress. Frontend styling has not been completed.

frontend/pages/entities/[slug].page.tsx
```diff
@@ -38,6 +49,25 @@ const EntityDetailPage: FC<EntityProps> = ({ entity }: EntityProps) => {
       />
       <ModuleContainer>
         <ModuleDetail module={entity} />
+        {entity.relatedEntities?.map((relatedEntity) => (
```
Comment (iacobfred) [inline frontend/pages/entities/[slug].page.tsx:52]: There probably should be a "Related entities" header/label above wherever the related entities are displayed (but only if `!!entity.relatedEntities?.length`).

iacobfred (changes requested):

aaw0151:
image

Current implementation layout.

frontend/pages/entities/[slug].page.tsx
```diff
@@ -21,6 +21,20 @@ const EntityRelatedQuoteCard = styled(Card)({
...
+  textOverflow: "ellipsis",
+  minHeight: "1rem",
+  marginBottom: "1rem",
+  width: "17rem",
+  color: "black",
+  "&:hover": {
+    backgroundColor: "#c4e1ff",
+  },
+});
+
```
Comment (mattfredericksen) [inline frontend/pages/entities/[slug].page.tsx:37]: Let's extract all the styles that are shared between this and `EntityRelatedQuoteCard` to a `sharedStyles` object.
E.g.
```ts
const sharedStyles: CSSProperties = {
  cursor: "pointer",
}

const EntityRelatedQuoteCard = styled(Card)({
  ...sharedStyles,
  otherStyles
})

const EntityRelatedEntityCard = styled(Card)({
  ...sharedStyles,
  otherStyles
})
```

frontend/pages/entities/[slug].page.tsx
```diff
@@ -38,6 +52,27 @@ const EntityDetailPage: FC<EntityProps> = ({ entity }: EntityProps) => {
...
+              {entity.relatedEntities?.map((relatedEntity) => (
+                <Link href={`/entities/${relatedEntity.slug}`} key={relatedEntity.slug} passHref>
+                  <a>
+                    <EntityRelatedEntityCard raised sx={{ margin: "0.3rem" }}>
+                      <Typography sx={{ padding: "0.5rem", textAlign: "center" }}>
+                        <HTMLEllipsis unsafeHTML={relatedEntity.name} maxLine="1" basedOn="words" />
+                      </Typography>
+                    </EntityRelatedEntityCard>
+                  </a>
+                </Link>
```
Comment (mattfredericksen) [inline frontend/pages/entities/[slug].page.tsx:71]: We don't need `passHref` here.
https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

mattfredericksen (changes requested):

PR #900, opened 2021-12-02
Refactoring

iacobfred:


core/models/model.py
```diff
@@ -254,19 +260,7 @@ def get_updated_placeholder(cls, match: Match) -> str:
...
-    model = serpy.MethodField()
-    absolute_url = serpy.StrField()
-
-    def get_model(self, instance: ExtendedModel) -> str:
-        """Return the model name of the instance."""
-        return get_model_name(instance)
-
-
-class DrfModelSerializer(ModeratedModelSerializer):
+class ModelSerializer(serializers.ModelSerializer):
```
mattfredericksen, line 263: Not sure if this is a problem

core/models/model.py
```diff
@@ -294,7 +288,8 @@ class Meta(ModeratedModelSerializer.Meta):
         fields = ['pk']
 
 
-class DrfTypedModelSerializerMixin(serializers.ModelSerializer):
+class TypedModelSerializerMixin(ModelSerializer):
+    """Base serializer for ModularHistory's typed models."""
```
mattfredericksen, line 292: Maybe affected by https://github.com/ModularHistory/modularhistory/pull/900/files#r760768310

mattfredericksen (approved):
It's ok I guess 🤷‍♂️

apps/collections/api/serializers.py
```diff
@@ -1,62 +1,26 @@
...
-from apps.quotes.models.quote import Quote
 from apps.sources.api.serializers import SourceSerializer
-from apps.sources.models.source import Source
 from core.models.serializers import SluggedModelSerializer
 
 
 class CollectionSerializer(SluggedModelSerializer):
     """Serializer for collection."""
 
-    propositions_serialized = PropositionSerializer(
```
alashow: We need both fields for this to work correctly. Non-serialized version accepts for writes with ids and serialized versions were being used as read-only. With these changes, it's now not working as expected because the api now expects whole sources, quote etc objects instead of existing ids.

iacobfred: @alashow, I'm confused because I don't see `propositions_serialized` being removed in the file diff for `apps/collections/api/serializers.py` at https://github.com/ModularHistory/modularhistory/pull/900/files

It looks to me like I just renamed the serializers without removing anything...

PR #908, opened 2021-12-03
MH-131: refactoring

iacobfred:


apps/moderation/serializers.py
```diff
@@ -51,8 +51,8 @@ def get_moderated_fields(self) -> list[dict]:
                     'name': field_name,
                     'type': field.__class__.__name__,
                     'editable': not getattr(field, 'read_only', False),
-                    'required': not getattr(field, 'required', False),
```
alashow, line 54: Where did these two `not`s come from? 🤔

iacobfred: No idea. Noticed them while discussing fields API behavior with @mattfredericksen.

core/settings.py
```diff
@@ -53,7 +53,7 @@
 # https://docs.djangoproject.com/en/dev/ref/settings#s-allowed-hosts
 ALLOWED_HOSTS = config(
     'ALLOWED_HOSTS',
-    default='localhost, 127.0.0.1, 0.0.0.0, django',
+    default='localhost, 127.0.0.1, django',
```
alashow, line 56: I hope this is intentional?

alashow (approved):
LGTM

PR #909, opened 2021-12-04
Tasks/serializer fixes

alashow:


apps/moderation/tests/test_serializers.py
```diff
@@ -0,0 +1,72 @@
...
+from .moderated_serializers import MODERATED_SERIALIZERS as serializers
+
+
+@pytest.mark.django_db()
+class TestModeratedSerializers:
+
+    # allow these fields to have choices instead of instant search
+    instant_search_field_exceptions = ['birth', 'death']
+
+    @classmethod
```
iacobfred, line 16: Why class methods?

apps/search/documents/entity.py
```diff
@@ -4,6 +4,7 @@
 from apps.entities.models.entity import Entity
 from apps.search.documents.config import DEFAULT_INDEX_SETTINGS, html_field_analyzer
 
+from ...entities.models import Category
```
iacobfred, line 7: nit: prefer absolute imports like `from apps.entities.models import Category`

apps/moderation/tests/test_serializers.py
alashow: IDE kept saying it should be classmethod since they're not using anything self. Should I change?

iacobfred: Gotcha. No need to change it )

apps/moderation/serializers.py
```diff
@@ -46,13 +59,14 @@ def get_moderated_fields(self) -> list[dict]:
...
                     'type': field.__class__.__name__,
-                    'editable': not getattr(field, 'read_only', False),
-                    'required': not getattr(field, 'required', False),
-                    'allow_blank': not getattr(field, 'allow_blank', False),
+                    'editable': True,  # TODO: remove after front-end code is updated
+                    'required': getattr(field, 'required', False),
+                    'allow_blank': getattr(field, 'allow_blank', False),
                     'verbose_name': getattr(field, 'verbose_name', None),
                     'help_text': getattr(field, 'help_text', None),
                     'choices': self.get_choices_for_field(field, field_name),
```
mattfredericksen, line 72: Why do we want `editable: True` here?

mattfredericksen (approved):
LGTM

alashow:
I assumed frontend might be using it?

mattfredericksen:
> I assumed frontend might be using it?

Oh, I meant why `editable: True` instead of `not getattr(field, 'read_only', False)`.

alashow:
> > I assumed frontend might be using it?
> 
> Oh, I meant why `editable: True` instead of `not getattr(field, 'read_only', False)`.

Because now we're filtering out non editable fields (with `if is_editable`)

PR #912, opened 2021-12-08
MH-44: add module to collection

iacobfred:


alashow (approved):
LGTM

apps/collections/api/views.py
```diff
@@ -12,3 +24,63 @@ class CollectionViewSet(ModelViewSet):
...
+    """Add items to the collection."""
+
+    def apply_items(self, collection: Collection, items: dict):
+        for module_type in COLLECTION_MODULE_TYPES:
+            try:
+                objects = items.get(module_type, [])
+                if objects:
+                    getattr(collection, module_type).add(*objects)
+            except IntegrityError:
+                raise ValidationError(f'Some {module_type} were not found in the database')
```
alashow, line 76: This could be moved to parent class in the future so CollectionRemoveItemsView can use similar validation logic too.

